PR #52, opened 2026-02-03
Implementing IfThen Expression Type To and From Substrait text

gord02:
## Description

This PR implements the ifThen Expression Type To and From Substrait text.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing
- Added test in in both expressions.rs files. And ran locally to confirm working parsing. 
- Will add proper roundtrip test in later PRs

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -358,17 +358,17 @@ Currently, only references to fields in the Relations' input are supported.
 #### Examples
 
 ```rust
-use substrait_explain::parser::Parser;
-
-let plan_text = r#"
+# use substrait_explain::parser::Parser;
+# 
+# let plan_text = r#"
 === Plan
 Root[result]
   Project[$0, $1, $42]
     Read[data => field0:i64, field1:string, field42:boolean]
-"#;
-
-let plan = Parser::parse(plan_text).unwrap();
-assert_eq!(plan.relations.len(), 1);
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
 ```
 
 ### Function Calls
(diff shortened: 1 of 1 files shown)


src/parser/expressions.rs
```diff
@@ -477,11 +477,87 @@ impl ScopedParsePair for Expression {
...
+    ) -> Result<Self, MessageParseError> {
+        assert_eq!(pair.as_rule(), Self::rule());
+
+        let mut iter = RuleIter::from(pair.into_inner()); // should have 2 or more children
+
+        let mut ifs: Vec<IfClause> = Vec::new();
+
+        // gets all of the if clauses
+        while let Some(p) = iter.try_pop(Rule::if_clause) {
+            let if_clause = Some(IfClause::parse_pair(extensions, p));
```
gord02: Some should not be used to wrap the results of parse_pair. Some(<X>) always produces a Some(), so the None block will never be reached. We should just match the result of parse_pair in the match. This prevents potentially panics in parsing resulting from the uncaught error from unwrapping the Some() created. We should also not use error propogation here(?) since that this terminates execution which we don't want when parsing.

src/textify/expressions.rs
```diff
@@ -690,6 +690,30 @@ impl Textify for ArgType {
...
+            let if_expr = ctx.expect(clause.r#if.as_ref());
+            let then_expr = ctx.expect(clause.then.as_ref());
+            write!(&mut output, "{if_expr} -> {then_expr}")?;
+            output.push_str(", ");
+        }
+
+        let else_expr = ctx.expect(self.r#else.as_deref());
+        write!(&mut output, "_ -> {else_expr}")?;
+        output.push(')');
+        write!(w, "{output}")
```
wackywendell: Instead of writing to a string, and then writing the string to `w`, it's more idiomatic (simpler code, and more performant) to just write to `w`:

```suggestion
        write!(w, "if_then(")?;
        for clause in &self.ifs {
            let if_expr = ctx.expect(clause.r#if.as_ref());
            let then_expr = ctx.expect(clause.then.as_ref());
            write!(w, "{if_expr} -> {then_expr}, ")?;
        }

        let else_expr = ctx.expect(self.r#else.as_deref());
        write!(w, "_ -> {else_expr})")
```

wackywendell (changes requested):
In general, good code change!

I have one moderate suggestion, see attached, and two significant ones:

1. Let's update `GRAMMAR.md` to note the addition of this to the grammar.
2. Let's also add or extend a roundtrip test to include an `IfThen` example. In general, testing should not come in separate PRs, but should come in the same PR as the implementation.

PR #64, opened 2026-02-26
Switch parser to Lalrpop

wackywendell:
This PR switches from Pest to lalrpop, to clean up the code substantially:

- No more `ParsePair`, `ScopedParsePair`, `RuleIter`, `unreachable!()`.
- There's an AST constructed from the grammar, with a second pass for 'lowering' from the AST to the protobuf message. This makes for a much more straightforward implementation.
- Unfortunately, the `line_grammar.lalrpop` is not quite as readable as the pest grammar, but it's not bad.
- Also adds some docs to modules in a number of places.
- It's a big PR, but it's -1700 lines overall.

See `adrs/lalrpop.md` to see the reasoning for this, `parser/line_grammar.lalrpop` for the grammar, and `parser/ast.rs` for the AST.

This is a draft PR because I'm a bit undecided still, and haven't fully reviewed it. This was all done via @codex partly in the cloud and partly locally.


wackywendell:
Closing this draft for now because it is too large a PR to review and land well as-is.

The part I still think is worth preserving is the separate lowering step: having an intermediate representation before producing Substrait proto messages seems useful independently of any parser-library change. If we come back to this, I think the better shape is to split that out first, then consider parser implementation changes separately.

PR #100, opened 2026-04-14
fix: use extension registry for CLI text parsing

### LINKED ISSUE #95: CLI text parser doesn't use extension registry
## Bug

The CLI's `convert -f text` and `validate` commands don't pass the `ExtensionRegistry` to the text parser, so plans containing custom extensions (registered via `run_with_extensions`) fail to parse from text input.

### Root cause

In `src/cli.rs`, `Format::read_plan` for `Format::Text` (line 374) calls `parse(&input_text)` without the registry:

```rust
Format::Text => {
    let input_text = read_text_input(reader)?;
    Ok(parse(&input_text)?)
}
```

Similarly, `run_validate_with_io` (line 245) calls `parse(&input_text)` directly.

Both should use the registry-aware parser (`Parser::new().with_extension_registry(registry).parse_plan(...)`) so that custom extensions are recognized.

### Impact

- `convert -f json -t text` **works** (JSON parsing uses descriptors, text formatting uses registry)
- `convert -f text -t json` **fails** with: `Unregistered extension 'ReadRelEnhancement'`
- `validate` **fails** with the same error

This means any binary that registers custom extensions via `run_with_extensions` can only convert *to* text, not *from* text.

### Expected behavior

All commands should use the extension registry for both parsing and formatting, so that round-tripping through the CLI works in all directions.

wackywendell:
## Description

The CLI's `convert -f text` and `validate` commands didn't pass the `ExtensionRegistry` to the text parser, so plans containing custom extensions (registered via `run_with_extensions`) failed to parse from text input.

This PR:
- Adds `parse_with_registry()` to `lib.rs` as the symmetric counterpart to `format_with_registry()`
- Fixes `Format::read_plan` for text input to use the registry-aware parser
- Refactors `run_validate_with_io` to reuse `Format::read_plan` / `write_plan` instead of duplicating the parse/format logic

## Type of Change

- [x] Bug fix

## Testing

- [x] All existing tests pass (176 unit + 85 integration/doc tests)
- [x] Pre-commit hooks pass (clippy, fmt)

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #95

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -7,7 +7,7 @@ use clap::{Parser, Subcommand};
 use prost::Message;
 
 use crate::extensions::ExtensionRegistry;
-use crate::{FormatError, OutputOptions, Visibility, format_with_registry, parse};
+use crate::{FormatError, OutputOptions, Visibility, format_with_registry, parse_with_registry};
 
 /// The outcome of a CLI operation.
 ///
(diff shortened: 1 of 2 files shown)


benbellick (approved):
Looks reasonable to me. Though I wonder if there is a way to test this. Maybe we could have some dummy protos inside of this repo for testing purposes?

PR #101, opened 2026-04-14
feat: support VirtualTable Read with inline and verbose text formats

### LINKED ISSUE #56: [FEATURE] Support Textify Representation of VirtualTable ReadRel Type
## Summary

Adds support for textify parsing of VirtualTable ReadRel from substrait json. 

## Proposed Solution

The following is an example of how the textified representation of VritualTables could look:
```
  VirtualRead[] rowCount=1
      Row[]
```

wackywendell:
## Description

Add support for VirtualTable in ReadRel (closes #56), enabling inline literal data tables in Substrait plans — similar to SQL `VALUES` clauses.

Two text formats are supported:

**Inline** (default for small tables, `rows * columns <= 8`):
```
Read:Virtual[(1, 'alice'), (2, 'bob') => id:i64, name:string]
```

**Verbose** (default for larger tables):
```
Read:Virtual[_ => id:i64, name:string]
  + Row[1, 'alice']
  + Row[2, 'bob']
```

Both forms are always accepted by the parser; the textifier selects based on a configurable threshold. Mixed form (inline rows + `+ Row` addenda) is accepted but never produced.

### Design decisions

- `Read:Virtual` uses the `Type:Subtype` colon pattern (consistent with `ExtensionLeaf:Name`)
- `+ Row` uses the `+`-prefixed addendum pattern (same structural role as `+ Enh:` / `+ Opt:`)
- `Addendum` enum unifies `AdvExt` and `Row` handling in the structural parser

### Parser refactoring (included)

VirtualTable support exposed an opportunity to make the parser trait more self-contained:

- **`RelationParsePair` redesign**: `parse_pair_with_context` now returns `(Self, usize)` — each relation type computes its own output field count during parsing rather than having it reconstructed afterward. `into_rel` takes `RelationAddenda` (pre-parsed rows + advanced extension) so each type handles its own addenda.
- **Identity emit**: new `make_emit` helper produces `EmitKind::Direct` for identity output mappings instead of an explicit `Emit`, matching Substrait semantics. This also fixes a Fetch textifier regression where passthrough columns were emitted as empty when `EmitKind::Direct` was in effect.
- **Structured error messages**: `ParseError::ValidationError` now carries `ParseContext` for line-precise diagnostics.
- Extension relations now reject `+ Enh:` / `+ Opt:` addenda with an explicit error (previously silently applied).

## Type of Change

- [x] New feature
- [x] Documentation update

## Testing

- [x] Added 8 roundtrip tests (inline, verbose, empty table, single-column, large table uses verbose, VirtualTable in complex plan, error paths)
- [x] All existing tests pass (309 total)
- [x] GRAMMAR.md doc tests compile and pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #56

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -527,6 +527,91 @@ Root[result2]
 # assert_eq!(plan.relations.len(), 2);
 ```
 
+### VirtualTable Read Relation
+
+A VirtualTable read embeds inline data directly in the plan, similar to SQL's `VALUES` clause. Instead of referencing a catalog table, the data rows are specified as part of the relation.
+
+The `Read:Virtual` relation uses the same `ReadRel` protobuf message with `ReadType::VirtualTable`, where each row is a `nested::Struct` containing expressions.
+
+#### Syntax
+
+There are two forms: **inline** (rows as arguments) and **verbose** (`+ Row` addenda).
+
+**Inline form:**
+
+`"Read:Virtual" "[" (virtual_row ("," virtual_row)*)? "=>" named_column_list "]"`
+
+Where `virtual_row := "(" expression ("," expression)* ")"` — a parenthesized tuple of expressions forming one row.
+
+**Verbose form:**
+
+```text
+"Read:Virtual" "[" "_" "=>" named_column_list "]"
+  ("+ Row" "[" expression ("," expression)* "]")*
+```
+
+The `+ Row` lines are addenda attached to the `Read:Virtual` relation (same structural role as `+ Enh:` / `+ Opt:` lines). They must appear before any child relations.
+
+Both forms are always accepted by the parser. The textifier uses inline form when `rows * columns <= 8`, and verbose form otherwise.
+
+**Mixed form (accepted, not recommended):** If a `Read:Virtual` has both inline rows and `+ Row` addenda, the inline rows come first, followed by the addendum rows. The textifier will never produce mixed form.
+
+#### Components
+
+- `virtual_row` - parenthesized tuple of expressions, one per row
+- `expression` - any expression (literal, field reference, function call)
+- `named_column_list` - output column names with type annotations
+- `_` - empty marker (no inline rows; rows provided via `+ Row` addenda)
+
+#### Examples
+
+Inline form with two rows:
+
+```rust
+# use substrait_explain::parser::Parser;
+#
+# let plan_text = r#"
+=== Plan
+Root[id, name]
+  Read:Virtual[(1, 'alice'), (2, 'bob') => id:i64, name:string]
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
+```
+
+Verbose form with `+ Row` addenda:
+
+```rust
+# use substrait_explain::parser::Parser;
+#
+# let plan_text = r#"
+=== Plan
+Root[id, name]
+  Read:Virtual[_ => id:i64, name:string]
+    + Row[1, 'alice']
+    + Row[2, 'bob']
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
+```
+
+Empty virtual table (no rows):
+
+```rust
+# use substrait_explain::parser::Parser;
+#
+# let plan_text = r#"
+=== Plan
+Root[id, name]
+  Read:Virtual[_ => id:i64, name:string]
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
+```
+
 ### Filter Relation
 
 #### Syntax
--- a/src/parser/relations.rs
+++ b/src/parser/relations.rs
@@ -15,9 +15,7 @@ use substrait::proto::{
     RelCommon, SortField, SortRel, Type, aggregate_rel, join_rel, read_rel, r#type,
 };
 
-use super::{
-    ErrorKind, MessageParseError, ParsePair, Rule, RuleIter, ScopedParsePair, unwrap_single_pair,
-};
+use super::{MessageParseError, ParsePair, Rule, RuleIter, ScopedParsePair, unwrap_single_pair};
 use crate::extensions::any::Any;
 use crate::extensions::registry::{ExtensionError, ExtensionType};
 use crate::extensions::{ExtensionArgs, ExtensionRegistry, SimpleExtensions};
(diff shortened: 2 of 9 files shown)


wackywendell:
@codex review

src/parser/relations.rs
```diff
@@ -350,6 +339,92 @@ impl RelationParsePair for ReadRel {
...
+        rows.push(nested::Struct {
+            fields: parse_expression_list(extensions, expression_list),
+        });
+    }
+
+    // TODO: Validate that each row has the same number of expressions as columns.
+    // Currently no parser-side warning mechanism exists (the parser is fail-fast,
+    // and a width mismatch isn't a syntax error). Consider adding once a warning
+    // collection path is available.
+    let column_count = columns.len();
```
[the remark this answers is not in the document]
wackywendell: Already considered — see the TODO comment right above this line (369-372). The parser is currently fail-fast with no warning path, and a width mismatch is more of a semantic validation issue than a syntax error. We'll address it when a warning collection mechanism is added.

src/parser/structural.rs
```diff
@@ -331,33 +368,32 @@ impl<'a> TreeBuilder<'a> {
...
                     return Err(ParseError::ValidationError(format!(
-                        "line {}: advanced extension annotations (+ Enh: / + Opt:) must \
+                        "line {line_no}: addenda (+ Enh: / + Opt: / + Row) must \
                          appear before child relations, not after",
-                        adv_ext.line_no
                     )));
                 }
 
-                parent.adv_extensions.push(adv_ext);
+                parent.addenda.push(addendum);
```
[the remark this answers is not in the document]
wackywendell: Good catch — this is actually a pre-existing issue (the old `adv_extensions` field had the same silent-discard behavior for `+ Enh:` / `+ Opt:` on Root), but extending addenda to include `+ Row` makes it worth fixing now. Added a validation check in `build_plan_rel` that rejects non-empty addenda on Root relations.

GRAMMAR.md
```diff
@@ -527,6 +527,91 @@ Root[result2]
...
+Where `virtual_row := "(" expression ("," expression)* ")"` — a parenthesized tuple of expressions forming one row.
+
+**Verbose form:**
+
+```text
+"Read:Virtual" "[" "_" "=>" named_column_list "]"
+  ("+ Row" "[" expression ("," expression)* "]")*
+```
+
+The `+ Row` lines are addenda attached to the `Read:Virtual` relation (same structural role as `+ Enh:` / `+ Opt:` lines). They must appear before any child relations.
```
benbellick, line 553: I think that the above short form is great, though I have some questions about the verbose form. 

I want to ask why `+ Row` has the same structural role as `+ Enh` and `+ Opt`. Both enhancements and optimizations are additions _on top of_ the core substrait, and the `+` communicates that it is an addition on top of a core. On the other hand, virtual tables are a core part of the spec. It would make sense to me for them to be visually different.


One other thing, it is counterintuitive to me that the inline format would use parentheses, but the verbose form would use brackets. Parenthesis communicate tuple-like things with a fixed width, so I think they are more appropriate for representing relations in both cases. 

What do you think about one of the following alternative representations? 

#### same as inline but with new lines
```
Root[id, name]
  Read:Virtual[_ => id:i64, name:string]
    (1, 'alice')
    (2, 'bob')
```


#### using dashes to demarcate new rows
```
Root[id, name]
  Read:Virtual[_ => id:i64, name:string]
    - (1, 'alice')
    - (2, 'bob')
```


And then one more item, it looks like there is precedent in other relations for not having the `=>` syntax in the brackets after the relation name. What do you think about just dropping the `_ =>` after the verbose form? 

E.g.
```
Root[id, name]
  Read:Virtual[id:i64, name:string]
    (1, 'alice')
    (2, 'bob')
```

src/textify/foundation.rs
benbellick, line 1: Not for this PR but future work: it is not clear to me from the name `foundation` alone what is meant to go in this file. I propose that we either:
- break this up into a few files with clearer names, or
- add a docstring here clarfying what is intended to go here.

src/textify/rels.rs
```diff
@@ -366,48 +378,99 @@ impl<'a> Textify for TableName<'a> {
...
+                    }),
+                    columns,
+                    emit,
+                    advanced_extension: rel.advanced_extension.as_ref(),
+                    rows: vec![],
+                    children: vec![],
+                }
+            }
+            Some(ReadType::VirtualTable(vt)) => {
+                let col_count = columns.len();
```
benbellick, line 403: just a suggestion, take it or leave it, but what do you think about extracting this out into a helper function? The nesting is getting pretty deep here.

benbellick (commented):
Thanks for this! I didn't look _too_ closely at the code because it may change based off of the discussion around the verbose form of the `VirtualTables`. One way we could move forwards though and make the PR smaller could be to open a new PR where we just handle the compact form, which I am fully in agreement with. Then after that we can tackle the verbose form separately. What do you think?

GRAMMAR.md
wackywendell: Thanks for the thoughtful feedback! I agree on splitting this. The inline form is ready now, and the verbose form deserves its own design discussion. I'll strip the verbose form out of this PR and file issues for the design questions you raised.

On the design questions:

### Why `+` for rows

In my mind, `+` means "this is not a new operator in the tree; it's additional data attached to the operator above." It's about tree structure — how lines relate to their parent — rather than core-vs-non-core Substrait. `+ Enh`, `+ Opt`, and `+ Row` all share the property that they modify a relation without introducing a new child in the plan hierarchy.

### Why `[...]` instead of `(...)`

The `+ <name>[...]` pattern is context-independent — the parser can identify and parse a `+ Row[...]` line from its first tokens alone, without knowing what relation it's attached to. Bare `(1, 'alice')` on an indented line would be the only indented content that isn't a relation or a `+ ...` addendum, breaking the invariant that line type is identifiable from the first token.

### On dropping `_ =>`:

I think `=>` should stay for the moment. Currently, `=>` is required on every relation that *can* have arguments; it's what disambiguates "these are arguments" from "these are output columns." `_ =>` means "I could have arguments here, I just don't". that's structurally different from Root and Project, which *never* take arguments and so never use `=>`. Dropping it would make VirtualTable the first relation where `name:type` column definitions appear inside `[...]` with no separator, blurring that visual pattern. That said, making `=>` optional more broadly is worth exploring: see #104; I rather like the idea.

### Using `- ` for rows

I actually quite like using `- ` as a lighter-weight prefix for "additional arguments" more generally. It could work for extension nodes with verbose argument lists too, not just VirtualTable rows. Filed #105 for that.

I'll update this PR to inline-only and we can iterate on the verbose form separately.

src/textify/foundation.rs
wackywendell: Mm, yes, good point!

wackywendell:
Closing in favor of #106 / #107.

Issue #104, opened 2026-04-17
feat: make '=>' optional when relation arguments are absent


wackywendell:
## Summary

Make `=>` optional in the text format for relations where arguments are absent.

## Problem

Currently, `=>` is required on every relation that _can_ have arguments, even when those arguments are empty. This leads to the `_` placeholder pattern (e.g., `Read:Virtual[_ => id:i64, name:string]` (see #101), `Aggregate[_ => sum($0)]`, `Fetch[_ => $0, $1]`). While `_ =>` is structurally unambiguous, it's visual noise when the argument side is empty.

## Goal

Allow omitting `_ =>` when there are no arguments, so that e.g.:

- `Read:Virtual[_ => id:i64, name:string]` can be written as `Read:Virtual[id:i64, name:string]`
- `Aggregate[_ => sum($0)]` can be written as `Aggregate[sum($0)]`
- `Fetch[_ => $0, $1]` can be written as `Fetch[$0, $1]`

The `=>` form would still be accepted (and required when arguments are present).

## Complexities

- **Ambiguity between arguments and output columns**: Without `=>`, the parser must distinguish whether bracket contents are arguments or output columns from syntax alone. For most relations this is feasible (e.g., `name:type` columns vs. expressions), but edge cases may exist — especially for extension relations where arguments and columns share syntax.
- **PEG ordered choice**: The Pest parser can try `args => cols` first and fall back to `cols`-only, but each relation's grammar rule needs individual attention to ensure no false matches.
- **Textifier policy**: Should the textifier omit `_ =>` when possible, or keep it for consistency? Likely omit for cleaner output, but this is a design choice.
- **Backward compatibility**: The `_ =>` form must remain valid for parsing.

## Success Criteria

- Relations with no arguments can omit `_ =>`
- `_ =>` is still accepted everywhere it currently works
- No ambiguity introduced for any relation type
- Textifier updated to emit the shorter form when arguments are absent
- Roundtrip tests pass

wackywendell:
Some additional complications:

- For "no output columns", we currently output `Operator[args => ]`, with the `=>` but no `_`. That seems inconsistent.
- For extension relations... they may not really have columns...

wackywendell:
Some of the complications above have been clarified by work in #106:

**Resolved** — "no output columns" now has a consistent canonical form:

[...]

PR #106, opened 2026-04-17
feat: support VirtualTable Read with inline format

### LINKED ISSUE #56: [FEATURE] Support Textify Representation of VirtualTable ReadRel Type

wackywendell:
## Description

> Depends on #107 (parser refactoring), which must merge first. Factored out of #101 - this is a simpler version of that (only inline parsing, no `+ Row` syntax).

Add support for `VirtualTable` in `ReadRel` (closes #56), enabling inline literal data tables in Substrait plans — similar to SQL `VALUES` clauses.

The virtual read is written as a `Read:Virtual` operator, with rows inline as positional arguments:

```
Read:Virtual[(1, 'alice'), (2, 'bob') => id:i64, name:string]
```

Use `_` for an empty virtual table (no rows):

```
Read:Virtual[_ => id:i64, name:string]
```

### Design decisions

- `Read:Virtual` uses the `Type:Subtype` colon pattern (consistent with `ExtensionLeaf:Name`)
- Uses the `Addendum` enum introduced in #107 for structural parser integration
- `GRAMMAR.md` updated with syntax, examples, and doc-tested code blocks

## Type of Change

- [x] New feature
- [x] Documentation update

## Testing

- [x] Added roundtrip tests (inline, empty table, single-column, VirtualTable in complex plans)
- [x] All existing tests pass
- [x] GRAMMAR.md doc tests compile and pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #56

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -537,14 +537,13 @@ The `Read:Virtual` relation uses the same `ReadRel` protobuf message with `ReadT
 
 `"Read:Virtual" "[" (virtual_row ("," virtual_row)*)? "=>" named_column_list "]"`
 
-Where `virtual_row := "(" expression ("," expression)* ")"` — a parenthesized tuple of expressions forming one row. Use `_` in place of rows for an empty virtual table.
+Where `virtual_row := "(" (expression ("," expression)*)? ")"` — a parenthesized tuple of expressions forming one row. Rows may be empty (`()`) for zero-column tables. For an empty virtual table, write `_` in place of the entire row list, e.g. `Read:Virtual[_ => id:i64]` for zero rows.
 
 #### Components
 
-- `virtual_row` - parenthesized tuple of expressions, one per row
+- `virtual_row` - parenthesized tuple of expressions, one per row; `()` for zero-column rows
 - `expression` - any expression (literal, field reference, function call)
 - `named_column_list` - output column names with type annotations
-- `_` - empty marker (no rows)
 
 #### Examples
 
@@ -758,9 +757,11 @@ positional_args := extension_arg ("," extension_arg)*
 extension_arg := reference / literal / expression
 named_args := named_arg ("," named_arg)*
 named_arg := name "=" extension_arg
-extension_columns := extension_column ("," extension_column)*
+extension_columns := (extension_column ("," extension_column)*)?
 extension_column := named_column / reference / expression
 ```
+
+Note: the parser also accepts the `=>` section being omitted entirely (e.g. `ExtensionLeaf:Foo[_]`), treating it as zero output columns. The canonical form always includes `=>`.
 
 #### Components
 
@@ -782,13 +783,13 @@ Root[result]
 Extension with positional arguments and no output columns:
 
 ```text
-ExtensionSingle:VectorNormalize[$0, $1, method='l2']
+ExtensionSingle:VectorNormalize[$0, $1, method='l2' => ]
 ```
 
 Extension with no arguments:
 
 ```text
-ExtensionLeaf:EmptySource[_]
+ExtensionLeaf:EmptySource[_ => ]
 ```
 
 #### Custom Extension Types
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -527,6 +527,57 @@ Root[result2]
 # assert_eq!(plan.relations.len(), 2);
 ```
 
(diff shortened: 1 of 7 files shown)


tests/plan_roundtrip.rs
```diff
@@ -468,6 +468,82 @@ Root[result]
...
+
+    roundtrip_plan(plan);
+}
+
+#[test]
+fn test_virtual_read_empty() {
+    let plan = r#"
+=== Plan
+Root[id, name]
+  Read:Virtual[_ => id:i64, name:string]"#;
```
benbellick, line 498: This is probably fine because I think this is the convention for other relations and it is better to be consistent, but I find the usage of `_` here a bit unintuitive. To me, `_` communicates "I don't care", which doesn't really make sense in this context, but that is where my mind goes. 

I think a more explicit syntax would be to use `[]` in all of the places, e.g.:
```
  Read:Virtual[[] => id:i64, name:string]"#;
```
or
```
  Read:Virtual[[(1), (2), (3)] => id:i64]"#;
```

tests/plan_roundtrip.rs
```diff
@@ -468,6 +468,82 @@ Root[result]
...
+=== Plan
+Root[result]
+  Read:Virtual[(unknown_func(1)) => result:i64]"#;
+
+    let result = Parser::parse(plan);
+    assert!(
+        result.is_err(),
+        "undeclared function in VirtualTable row should fail"
+    );
+}
```
benbellick, line 556: Are we able to handle degenerate tables which have no columns? 
E.g.
```
Root[result]
  Read:Virtual[(), () => ]"#;
```

GRAMMAR.md
```diff
@@ -527,6 +527,57 @@ Root[result2]
...
+
+A VirtualTable read embeds inline data directly in the plan, similar to SQL's `VALUES` clause. Instead of referencing a catalog table, the data rows are specified as part of the relation.
+
+The `Read:Virtual` relation uses the same `ReadRel` protobuf message with `ReadType::VirtualTable`, where each row is a `nested::Struct` containing expressions.
+
+#### Syntax
+
+`"Read:Virtual" "[" (virtual_row ("," virtual_row)*)? "=>" named_column_list "]"`
+
+Where `virtual_row := "(" expression ("," expression)* ")"` — a parenthesized tuple of expressions forming one row. Use `_` in place of rows for an empty virtual table.
```
benbellick:
```suggestion
Where `virtual_row := "(" expression ("," expression)* ")"` — a parenthesized tuple of expressions forming one row. For an empty virtual table, write `_` in place of the entire row list, e.g. `Read:Virtual[_ => id:i64]`. The `_` replaces the whole list and cannot appear alongside rows.
```

What do you think of something like this?

benbellick (approved):
Few comments but otherwise LGTM

tests/plan_roundtrip.rs
wackywendell: I see your point about `_`; it's used to mean 'unspecified value' in several languages, whereas here it means 'empty', which is quite different.

That said - I'm not really sure what would be a better replacement:

- **`[]` brackets around `args`**: I'd rather not add extra brackets if we don't need to - the syntax `operator[args... => columns]` seems fairly unambiguous, and having to put extra brackets around args everywhere adds clutter.
- **`()` for empty**: Nice syntax, but ambiguous: it's not clear if `Operator[() => …]` means "1 argument, an empty tuple" or "no arguments".
- **`empty` keyword**: E.g. `Operator[empty => …]`. Right now, there are no keywords, and this would conflict with the possibility of having an identifier - e.g. `Read[empty => …]` currently means "a read from the table named `empty`"
- **Nothing**: Allow either `Operator[columns]` like `Project` and `Root`(#104), or `Operator[=> columns]` / `Operator[args =>]` for empty columns / args. The first has precedent and should be OK; the second is unambiguous.

So for now, I think we keep it...

tests/plan_roundtrip.rs
wackywendell: Good point, added test and fixed 👍

PR #107, opened 2026-04-20
refactor: self-contained RelationParsePair with identity emit

wackywendell:
## Description

This PR refactors a bit to improve emit handling and advanced extension handling, by bringing them into the `RelationParsePair` trait - so that each `Relation` is responsible for both, rather than external handling.

This is factored out of #101, and a precedent for #106.

### Specifics

- **`RelationParsePair` redesign**: The output field count is now computed during parsing and returned from `parse_pair_with_context`, so it's part of the trait rather than computed externally. Similarly, `into_rel` takes `Option<AdvancedExtension>`, so it returns a fully complete `Rel`.
- **Identity emit**: new `make_emit` helper produces `EmitKind::Direct` for identity output mappings instead of an explicit `Emit`, matching Substrait semantics. This also fixes a `Fetch` textifier regression where passthrough columns were emitted as empty when `EmitKind::Direct` was in effect.
- **Structured error messages**: `ParseError::ValidationError` now carries `ParseContext` for line-precise diagnostics.
- Extension relations now reject `+ Enh:` / `+ Opt:` addenda with an explicit error (previously silently applied).

## Type of Change

- [x] Refactor

## Testing

- [x] All existing tests pass

Added 7 roundtrip tests exercising emit/output-field-count propagation across every relation type (`Read`, `Filter`, `Sort`, `Fetch`, `Join`, `Aggregate`) stacked under a `Project` with an expression, ensuring the expression's emit index is computed correctly.

## Checklist

- [x] Pre-commit hooks pass

### DIFF
--- a/src/parser/errors.rs
+++ b/src/parser/errors.rs
@@ -53,7 +53,7 @@ pub enum ParseError {
     #[error("Unknown extension relation type: {0}")]
     UnknownExtensionRelationType(String),
 
-    #[error("Invalid input at line {0}: {1}")]
+    #[error("Invalid input at {0}: {1}")]
     ValidationError(ParseContext, String),
 
     #[error("Error parsing section header on line {0}: {1}")]
--- a/src/parser/errors.rs
+++ b/src/parser/errors.rs
@@ -53,8 +53,8 @@ pub enum ParseError {
     #[error("Unknown extension relation type: {0}")]
     UnknownExtensionRelationType(String),
 
-    #[error("Validation error: {0}")]
-    ValidationError(String),
+    #[error("Invalid input at line {0}: {1}")]
+    ValidationError(ParseContext, String),
 
     #[error("Error parsing section header on line {0}: {1}")]
     Initial(ParseContext, #[source] MessageParseError),
--- a/src/parser/relations.rs
+++ b/src/parser/relations.rs
@@ -230,6 +230,14 @@ fn make_emit(output_mapping: Vec<i32>, direct_output_count: usize) -> EmitKind {
     }
 }
 
+/// Parse a reference list into an emit and output field count.
+fn parse_emit(reference_list: Pair<Rule>, direct_output_count: usize) -> (EmitKind, usize) {
+    let output_mapping = parse_output_mapping(reference_list);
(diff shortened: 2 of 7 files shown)


src/parser/errors.rs
```diff
@@ -53,8 +53,8 @@ pub enum ParseError {
     #[error("Unknown extension relation type: {0}")]
     UnknownExtensionRelationType(String),
 
-    #[error("Validation error: {0}")]
-    ValidationError(String),
+    #[error("Invalid input at line {0}: {1}")]
```
benbellick: Earlier in this file, there is:
```rust
impl std::fmt::Display for ParseContext {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "line {}: '{}'", self.line_no, self.line)
    }
}
```

So I think if you tried to render this error, it would look something like `Invalid input at line line` (double line).

src/parser/relations.rs
```diff
@@ -482,7 +498,7 @@ impl RelationParsePair for AggregateRel {
         pair: Pair<Rule>,
         input_children: Vec<Box<Rel>>,
         _input_field_count: usize,
```
benbellick, line 508: Since all of the other relations were updated to use `input_field_count`, it may be worth a comment why _this_ one doesn't need it.

src/parser/relations.rs
```diff
@@ -369,31 +380,33 @@ impl RelationParsePair for FilterRel {
...
-        // condition
         let condition = iter.parse_next_scoped::<Expression>(extensions)?;
         // references (which become the emit)
         let references_pair = iter.pop(Rule::reference_list);
         iter.done();
 
-        let emit = parse_reference_emit(references_pair);
+        let output_mapping = parse_output_mapping(references_pair);
+        let output_count = output_mapping.len();
+        let emit = make_emit(output_mapping, input_field_count);
```
benbellick: this group of three lines repeats 4 times. May be worth a helper function?

benbellick (approved):
some minor comments but otherwise LGTM! Nice refactor 🚀

PR #117, opened 2026-04-24
adding support for tuples as extension arguments

### LINKED ISSUE #116: [FEATURE] Implement Tuple for Extension Arguments.
## Summary

We should be able to have a list of arguments stored as a tuple in extension arguments. 

Example:
```
=== Plan
Root[result]
  Read[my.table => col:i64]
    + Enh:TupleSortHint[(&ASC, &DESC)]
```

gord02:
## Description

Adding support for tuples as extension arguments. 

## Related Issues

Closes #116

### DIFF
--- a/src/extensions/args.rs
+++ b/src/extensions/args.rs
@@ -153,6 +153,69 @@ impl Drop for ArgsExtractor<'_> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct TupleValue(Vec<ExtensionValue>);
+
+impl TupleValue {
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn iter(&self) -> std::slice::Iter<'_, ExtensionValue> {
+        self.0.iter()
+    }
+}
+
+impl fmt::Display for TupleValue {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "(")?;
+        for (i, item) in self.0.iter().enumerate() {
+            if i > 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "{item}")?;
+        }
+        if self.0.len() == 1 {
+            write!(f, ",")?;
+        }
+        write!(f, ")")
+    }
+}
+
+impl<'a> IntoIterator for &'a TupleValue {
+    type Item = &'a ExtensionValue;
+    type IntoIter = std::slice::Iter<'a, ExtensionValue>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
+impl IntoIterator for TupleValue {
+    type Item = ExtensionValue;
+    type IntoIter = std::vec::IntoIter<ExtensionValue>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl FromIterator<ExtensionValue> for TupleValue {
+    fn from_iter<I: IntoIterator<Item = ExtensionValue>>(iter: I) -> Self {
+        TupleValue(iter.into_iter().collect())
+    }
+}
+
+impl From<Vec<ExtensionValue>> for TupleValue {
+    fn from(items: Vec<ExtensionValue>) -> Self {
+        TupleValue(items)
+    }
+}
+
 /// Represents a value in extension arguments
 #[derive(Debug, Clone)]
 pub enum ExtensionValue {
@@ -169,7 +232,7 @@ pub enum ExtensionValue {
     /// Enum value (e.g. &CORE, &Inner) — Uses the wrapper EnumValue. the string holds the identifier without the `&` prefix
     Enum(String),
     /// Tuple of values, e.g. (&HASH, &RANGE) or (42, 'hello')
-    Tuple(Vec<ExtensionValue>),
+    Tuple(TupleValue),
     /// Expression (function call, etc.) — not yet fully supported, hence the
     /// private interface.
     #[allow(private_interfaces)]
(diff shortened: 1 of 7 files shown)


src/parser/expression_grammar.pest
```diff
@@ -308,16 +297,18 @@ extension_name = { extension_type ~ ":" ~ name }
 extension_type = { "ExtensionLeaf" | "ExtensionSingle" | "ExtensionMulti" }
 
 arguments = { (empty | (extension_arguments ~ (sp ~ "," ~ sp ~ extension_named_arguments)?) | extension_named_arguments)? }
-// Positional arguments (expressions, literals, references, enums)
+// Positional arguments (expressions, literals, references, enums, tuples)
 extension_arguments = { extension_argument ~ (sp ~ "," ~ sp ~ extension_argument)* }
-extension_argument  = { enum_value | reference | literal | expression }
+// tuple must come after expression due to overlapping grammar definitions
```
wackywendell: Can you clarify - what's overlapping with what here?

src/parser/expression_grammar.pest
```diff
@@ -308,16 +297,18 @@ extension_name = { extension_type ~ ":" ~ name }
 extension_type = { "ExtensionLeaf" | "ExtensionSingle" | "ExtensionMulti" }
 
 arguments = { (empty | (extension_arguments ~ (sp ~ "," ~ sp ~ extension_named_arguments)?) | extension_named_arguments)? }
-// Positional arguments (expressions, literals, references, enums)
+// Positional arguments (expressions, literals, references, enums, tuples)
 extension_arguments = { extension_argument ~ (sp ~ "," ~ sp ~ extension_argument)* }
-extension_argument  = { enum_value | reference | literal | expression }
+// tuple must come after expression due to overlapping grammar definitions 
+tuple = { "(" ~ sp ~ (extension_argument ~ (sp ~ "," ~ sp ~ extension_argument)*)? ~ sp ~ ")" }
```
wackywendell: If we're going to add tuples, we should probably do two things:

1. Ensure that 0-tuples and 1-tuples are writeable - i.e., how do I say "a tuple with zero elements" or "a tuple with exactly one element"?
2. The 1-tuple case needs to be distinguishable from an expression, where you can use parentheses around it without changing its meaning.

The standard for this in Python and Rust is the trailing comma convention:
- a 0-tuple is empty parentheses (`()`)
- a 1-tuple has the item followed by a comma (`(a,)`), with the comma showing that its a tuple and not an expression
- A multi-tuple uses comma separators (`(a, b)`, `(a, b, c)`), with an optional trailing comma.

Note that this also means that `(x)` is no longer parseable as a tuple, which I think is a good thing.

```suggestion
tuple = {
        // empty tuple (0-tuple)
        "(" ~ sp ~ ")"
        // Single element (1-tuple) - uses trailing comma to distinguish from expression
        | "(" ~ sp ~ extension_argument ~ sp ~ "," ~ sp ~ ")"
        // Multi-element tuple. Trailing comma optional.
        | "(" ~ sp ~ extension_argument
              ~ (sp ~ "," ~ sp ~ extension_argument)+
              ~ (sp ~ ",")? ~ sp ~ ")"
        }
```

src/extensions/args.rs
```diff
@@ -230,6 +242,22 @@ impl<'a> TryFrom<&'a ExtensionValue> for EnumValue {
     }
 }
 
+/// Helper for extracting the elements from an [`ExtensionValue::Tuple`].
+pub struct TupleValue(pub Vec<ExtensionValue>);
```
wackywendell: Two thoughts here:

1. This comment is confusing - who is this helping? Should users use this? When?
2. Having this exist but not really be used like this is a bit confusing. This type isn't carrying its weight.

I have two suggestions:

### Make `TupleValue` a first-class type

1. Change it to `ExtensionValue::Tuple(TupleValue)`. Then the `TupleValue` becomes a first-class object.
2. Make `items: Vec<ExtensionValue>` a private attribute.
3. Implement `IntoIterator` and `FromIterator` for `TupleValue`, so it can be iterated over directly; optionally also `From<Vec<ExtensionValue>>`.

This makes the `Vec` part an implementation detail, and allows us to e.g. implement `Display` / `Textify` directly on `TupleValue` which is a nice refactor.

### Drop `ExtensionValue`

We could just implement `TryFrom<ExtensionValue> for Vec<ExtensionValue>` and vice-versa; no newtype needed.

### Summary

I think I prefer the first approach - make it a first-class value. It adds a bit of boiler-plate (all those trait implementations...), but those are pretty straightforward. What do you think?

wackywendell (changes requested):
I have a couple comments here about the grammar (0-tuples and 1-tuples) and the structure (`TupleValue`); we should address those.

Additionally - we should also update `GRAMMAR.md` to reflect, and include round-trip tests for 0-tuples and 1-tuples.

src/parser/expression_grammar.pest
gord02: The tuple grammar overlaps with the cast expression grammar. if it came first, a cast expression like (78:i32)::i64 would have its (78:i32) matched as a one-element(literal) tuple, leaving ::i64 dangling. I added a comment with an example.

src/extensions/args.rs
gord02: I followed the first suggestion of making TupleValue first class. I also added the roundtrip tests for tuples of size 0, 1 and more as well as followed the rust/python convention for tuple lists.

src/parser/expression_grammar.pest
```diff
@@ -308,16 +297,31 @@ extension_name = { extension_type ~ ":" ~ name }
 extension_type = { "ExtensionLeaf" | "ExtensionSingle" | "ExtensionMulti" }
 
 arguments = { (empty | (extension_arguments ~ (sp ~ "," ~ sp ~ extension_named_arguments)?) | extension_named_arguments)? }
-// Positional arguments (expressions, literals, references, enums)
+// Positional arguments (expressions, literals, references, enums, tuples)
 extension_arguments = { extension_argument ~ (sp ~ "," ~ sp ~ extension_argument)* }
-extension_argument  = { enum_value | reference | literal | expression }
+
+// tuple must come after expression in extension_argument because cast_expression starts with "("
+// and must be tried first — otherwise "(78:i32)" would be consumed as a 1-element tuple.
```
wackywendell: > otherwise "(78:i32)" would be consumed as a 1-element tuple.

Is that still true? I think with the grammar below, it no longer is?

wackywendell (approved):
LGTM! Small comment, but good to go!

src/parser/expression_grammar.pest
gord02: Right. Yes. It is fixed now. Will remove the comment and merge.

PR #126, opened 2026-05-13
Add Timestamp and Duration conversions

### LINKED ISSUE #124: Expose timestamp and duration literal conversions for extension arguments
## Context

While reviewing DataDog/dd-source#427490, a downstream extension needed to represent timestamp and duration values in extension arguments. Today the practical carrier is usually `ExtensionValue::String`, but the semantic conversion is not just a string convenience: it needs precision-preserving formatting, strict parsing, and useful errors.

`substrait-explain` already has internal timestamp literal parsing/formatting for expressions. Extension authors would benefit from a public equivalent for extension argument values so they do not reimplement timestamp or duration handling locally.

## Proposal

Expose public helpers or wrapper types for timestamp and duration extension literals.

Possible shape:

- `TimestampLiteral::try_from(&ExtensionValue)` for parsing an extension string value.
- `TimestampLiteral::to_extension_value()` or a fallible conversion to `ExtensionValue::String` for formatting.
- Equivalent `DurationLiteral` support.
- Preserve sub-second precision and make timezone normalization explicit.
- Return `ExtensionError` with enough context for extension parse failures.

I would avoid blanket `From<Timestamp> for ExtensionValue` because timestamp formatting can fail for invalid or out-of-range protobuf values. A fallible API is more honest.

## Acceptance criteria

- Extension implementations can round-trip timestamp and duration argument values without custom chrono / protobuf JSON glue.
- Invalid timestamp or duration inputs fail at the extension parse boundary instead of silently normalizing or truncating.
- The public API makes it clear that these values are semantic literals carried by `ExtensionValue::String`, not arbitrary strings.

wackywendell:
## Description

Add support for converting between `ExtensionValue` and protobuf `Timestamp` and `Duration` types. This enables proper serialization and deserialization of these common protobuf types in extension arguments, with RFC3339 format for timestamps and protobuf JSON duration syntax for durations.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [ ] No breaking changes (or breaking changes documented)
- [x] Pre-commit hooks pass

## Related Issues

Closes #124

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/40f51889-83c1-4988-bfd9-91f4a129a052)

Comment @datadog to request changes

### DIFF
--- a/src/extensions/args.rs
+++ b/src/extensions/args.rs
@@ -6,7 +6,9 @@
 use std::collections::HashSet;
 use std::fmt;
 
+use chrono::{DateTime, Utc};
 use indexmap::IndexMap;
+use prost_types::{Duration as ProtoDuration, Timestamp};
 
 use super::ExtensionError;
 use crate::textify::expressions::Reference;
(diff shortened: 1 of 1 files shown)


wackywendell:
From discussion with BitsAI:

Wendell:
> Hm... zooming out, I think we want literals to work for extensions like they do elsewhere; so e.g. something like `'2026-05-13T15:00:00Z':timestamp` works for both. Does that work for regular literals? How might it work for extensions, structurally?

BitsAI:
> Short answer: yes for regular literals, not fully for extensions today.
>
> Regular expressions support typed literals like '2026-05-13T15:00:00Z':timestamp with semantic parsing in [src/parser/expressions.rs](https://github.com/datadog/substrait-explain/blob/dade3187e64bc3558c553320a08995eaefe35b2e/src/parser/expressions.rs), and textify preserves type suffixes via [src/textify/expressions.rs](https://github.com/datadog/substrait-explain/blob/dade3187e64bc3558c553320a08995eaefe35b2e/src/textify/expressions.rs). Extensions currently accept literal syntax in grammar ([src/parser/expression_grammar.pest](https://github.com/datadog/substrait-explain/blob/dade3187e64bc3558c553320a08995eaefe35b2e/src/parser/expression_grammar.pest)), but ExtensionValue parsing collapses it to primitive variants and drops the explicit type suffix in [src/parser/extensions.rs](https://github.com/datadog/substrait-explain/blob/dade3187e64bc3558c553320a08995eaefe35b2e/src/parser/extensions.rs). Textify then emits just '...' / number / bool in [src/textify/extensions.rs](https://github.com/datadog/substrait-explain/blob/dade3187e64bc3558c553320a08995eaefe35b2e/src/textify/extensions.rs).
So your example likely parses in extensions, but won’t round-trip with :timestamp intact.
>
> Best structural direction (minimal + review-friendly):
>
> 1. Add an extension literal representation that preserves optional type annotation (instead of dropping it).
>
> 2. Parse Rule::literal into that representation in extension parsing.
> 
> 3. Textify it back with suffix, matching regular literal formatting rules.
> 
> 4. Keep try_to_timestamp / try_to_duration as consumer helpers on top.
> 
> That keeps scope local to extension parsing/textify, avoids redoing chrono/prost logic, and preserves round-trip semantics.

So I think that general direction of supporting `'2026-05-13T15:00:00Z':timestamp` equivalently for the main code and the extension code is the right direction; not sure if the specific plan above is right.

PR #128, opened 2026-05-18
refactor(extensions)!: simplify extension argument handling

wackywendell:
## Description

This PR refactors `ExtensionArgs`, moving `ExtensionRelationType` out of `ExtensionArgs` and into `extensions`. Extension implementations now return only the values written inside their brackets: positional args, named args, and output columns.

This also clears up some poor parsing: previous to this PR, extension authors could specify the `ExtensionRelationType`, but it was ignored.

## Changes

- Add internal addendum labels for `+ Enh:` and `+ Opt:` so text syntax labels are separate from registry namespaces.
- Move extension relation form handling into parser-side relation parsing.
- Remove `ExtensionRelationType` from the public extension argument API.
- Replace `ExtensionArgs::new(...)` with `ExtensionArgs::default()` in examples, tests, and docs.

## API Impact

- Breaking: `ExtensionRelationType` and `ExtensionArgs::new(...)` are removed.
- Extension authors should construct `ExtensionArgs` with `ExtensionArgs::default()`.
- Extension authors no longer specify `Leaf`, `Single`, or `Multi` from `to_args()`.

## Validation

- `cargo test`
- `cargo test --doc`
- `cargo test --features=cli`
- `just check`

### DIFF
--- a/src/extensions/args.rs
+++ b/src/extensions/args.rs
@@ -37,6 +37,26 @@ use substrait::proto::expression::{RexType, reference_segment};
 use super::ExtensionError;
 use crate::textify::expressions::Reference;
 
+/// Kind of relation addendum in the text format.
+///
+/// Addenda are `+`-prefixed lines attached to relations. They are syntax-level
+/// constructs, distinct from [`crate::extensions::registry::ExtensionType`],
+/// which describes registry namespaces.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AddendumKind {
+    Enhancement,
+    Optimization,
+}
+
+impl AddendumKind {
+    pub(crate) fn prefix(self) -> &'static str {
+        match self {
+            AddendumKind::Enhancement => "Enh",
+            AddendumKind::Optimization => "Opt",
+        }
+    }
+}
+
 /// A Substrait expression carried as an extension argument or output column.
 ///
 /// Boxed because `proto::Expression` is large (multiple `Vec` fields in
--- a/src/extensions/args.rs
+++ b/src/extensions/args.rs
@@ -173,32 +173,20 @@ impl From<&str> for Expr {
     }
 }
 
-/// Represents text-format arguments for a registered advanced extension.
+/// Represents extension arguments plus optional output columns.
 ///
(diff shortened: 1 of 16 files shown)


src/extensions/args.rs
```diff
@@ -37,6 +37,26 @@ use substrait::proto::expression::{RexType, reference_segment};
 use super::ExtensionError;
 use crate::textify::expressions::Reference;
 
+/// Kind of relation addendum in the text format.
+///
+/// Addenda are `+`-prefixed lines attached to relations. They are syntax-level
+/// constructs, distinct from [`crate::extensions::registry::ExtensionType`],
+/// which describes registry namespaces.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AddendumKind {
```
gord02, line 46: nit: is this a good name? Maybe `AdvancedExtensionKind` or `AdvExtKind` instead?

src/parser/extensions.rs
```diff
@@ -358,12 +357,95 @@ impl ScopedParsePair for ExtensionColumn {
...
+pub(crate) enum ExtensionRelationKind {
+    Leaf,
+    Single,
+    Multi,
+}
+
+impl FromStr for ExtensionRelationKind {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
gord02, line 372: we no longer have the `as_str` is this expected?

gord02 (approved):

src/extensions/args.rs
wackywendell: `AddendumKind`is intentional here: this is not going to remain for only Advanced Extensions, as I'm about to add support for `ExtensionTable` with a `+ Ext` Addendum.

src/parser/extensions.rs
wackywendell: Yes, expected. `ExtensionRelationKind` is now parser-side only: it maps `ExtensionLeaf` / `ExtensionSingle` / `ExtensionMulti` text prefixes into the protobuf shape and is not used on the textify path, so there is no call site for `as_str` anymore.

PR #140, opened 2026-05-28
refactor!(api): privatize parser and textifier internals

wackywendell:
## Summary

- Make `parser`, `textify`, and crate fixtures private while preserving the crate-root public API.
- Keep extension fixtures available only where they are used by crate tests, and copy integration-test fixtures into `tests/common`.
- Update docs/doctests that previously referenced private fixtures or module paths.
- Clean up the parser child accumulator to move `Rel` values into protobuf boxes only at the generated-field boundary.

Using `cargo public-api -p substrait-explain --omit 'blanket-impls,auto-trait-impls,auto-derived-impls' | wc -l`, this reduces the public API from 1605 to 753 items.

## Review guide

This is the PR where the API boundary actually closes. The supported parser/formatter surface is now at the crate root: `Parser`, `parse`, `format`, `format_with_options`, `format_with_registry`, `OutputOptions`, parse/format errors, and the extension-authoring types under `extensions`.

`PlanWriter` stays private on purpose. The current type is generic over internal error accumulation and exposes formatter-phase methods such as `scope`, `write_extensions`, and `write_relations`. Re-exporting it would either expose more textifier internals or create a public writer API around implementation details. If we need a public writer later, it should be a smaller wrapper designed around stable inputs and outputs.

`extensions::examples` remains `#[doc(hidden)] pub` so doctests and examples can compile as external crates while using real extension handlers. It is crate-owned example support, not a stable extension API.

## Stack

1. #138
2. #139
3. This PR

## Validation

- `just test`
- `just check`
- `cargo clippy --examples --tests --all-features -- -D warnings`

### DIFF
--- a/src/textify/mod.rs
+++ b/src/textify/mod.rs
@@ -1,14 +1,13 @@
 //! Output a plan in text format.
 
 mod addenda;
-pub mod expressions;
-pub mod extensions;
-pub mod foundation;
-pub mod plan;
-pub mod rels;
-pub mod types;
+pub(crate) mod expressions;
+pub(crate) mod extensions;
+pub(crate) mod foundation;
+pub(crate) mod plan;
+pub(crate) mod rels;
+pub(crate) mod types;
 
-pub use foundation::{
-    ErrorQueue, FormatErrorType, OutputOptions, PlanError, Scope, ScopedContext, Textify,
-    Visibility,
-};
+#[cfg(test)]
+pub(crate) use foundation::ErrorQueue;
+pub(crate) use foundation::{OutputOptions, PlanError, Scope, ScopedContext, Textify, Visibility};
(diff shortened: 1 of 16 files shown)


src/textify/mod.rs
```diff
@@ -1,14 +1,13 @@
...
+pub(crate) mod plan;
+pub(crate) mod rels;
+pub(crate) mod types;
 
-pub use foundation::{
-    ErrorQueue, FormatErrorType, OutputOptions, PlanError, Scope, ScopedContext, Textify,
-    Visibility,
-};
+#[cfg(test)]
+pub(crate) use foundation::ErrorQueue;
```
gord02, line 12: nit: do these need to be two separate lines?

gord02 (approved):

src/textify/mod.rs
wackywendell: Technically, that would be legal - but its more conventional in Rust to leave `#[…]` on its own line. 🤔  `rustfmt` may even enforce it, not sure...

PR #141, opened 2026-05-29
Gordon.hamilton/update substrait version

gord02:
Upgrading the substrait version used by substrait-explain. I updated the code based on the resulting breaking changes.


wackywendell:
Note to self / other reviewers: the update of substrait-rs necessitated a few code changes:

1. Removing any support for URIs, as they have been removed from substrait
2. Correct handling of `RootReference` in `FieldIndex` (which could have been done separately?), with correct `unimplemented` handling of other references
3. `#[allow(deprecated)]` support for `Time` / `Timestamp`; should not affect users
4. `unimplemented` "support" for lambda expressions / references

wackywendell (approved):
LGTM!

PR #150, opened 2026-06-08
fix: zero arg textification and parsing fix

### LINKED ISSUE #144: [BUG] Function Signature Proto
## Description

In Substrait's compound function name format, the colon separates the function name from its argument type list. `row_number:` is a correct compound name for a zero-argument function — the colon is present but the argument type list is empty. Some substrait plans will use this format which does not roundtrip correctly when going back to proto plans. 

## Minimal Reproducible Example 
Example substrait json plan:
```
{
    "extensionFunction": {
      "functionAnchor": 1,
      "name": "count:",
      "extensionUrnReference": 1
    }
  }
```
Which is textified to:
```
=== Extensions
URNs:
  @  1: extension:io.substrait:functions_aggregate_generic
Functions:
  #  1 @  1: count:
```
Which when roundtripped:
tries to match rule on function name: `count:`
  
compound_name = @{ identifier ~ (":" ~ identifier)? } requiring a non-empty identifier after : → stops at count, leaving `:` unconsumed, so only count is registered as compound_name                                                                                     
 We would then try: `assert_eq!(pairs.as_str(), s)` where we get a mismatch,  ` assert_eq!(pairs.as_str(), s)` is just `count`, s is `count:`

## Suggested Fix
We should remove the colon suffix of the name if not followed by arguments to allow for proper round-tripping.

gord02:
## Description

Substrait encodes function names in extension declarations as compound names — a base name and a type-signature suffix joined by a colon (e.g. "add:i64_i64"). For zero-argument functions, the signature is empty, so the proto contains a trailing colon with nothing after it: "count:".

This form was not roundtrippable through the text format:

1. Proto → text: the text formatter wrote count: verbatim into the Extensions section
2. Text → proto: the grammar's compound_name rule requires an identifier after the colon, so count: failed to parse

We also wanted to retain support for functions without a signature(no colon to declare the list of arguments). 

## Type of Change

- [x] Bug fix

## Testing

- [x] All existing tests pass

## Related Issues
Closes #144

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -366,15 +366,41 @@ Root[result]
 
 #### Syntax
 
-`function_call := name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
+`function_call := compound_name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
+
+`compound_name := identifier (":" identifier?)?`
 
 #### Components
 
-- `name` - function name
+- `compound_name` - function name, optionally including a type-signature suffix (see below)
 - `anchor` - optional anchor (e.g., `#10`)
 - `urn_anchor` - optional URN anchor (e.g., `@1`)
 - `expression` - as above
 - `type` - optional output type
+
+#### Function Name Resolution
+
+A compound name is either *simple* (no colon, e.g. `add`) or *full* (has a colon, e.g. `add:i64_i64` or `count:`). These resolve differently:
+
+- **Simple** (`add`): matches any registered function whose base name is `add`, regardless of type signature. Succeeds only if the base name is unambiguous (exactly one registered function has that base name).
+- **Full** (`add:i64_i64`, `count:`): exact match only — the written name must match the registered name exactly.
+
+An `anchor` explicitly identifies which registered function is meant. When present:
+- A simple name validates that the anchor's registered name shares the same base.
+- A full name validates that the anchor's registered name matches exactly.
+
+The type signature and anchor are each optional, but either or both may be required to make the reference unambiguous.
+
+#### Examples
+
+```text
+add($0, $1)              // Simple: resolves if "add" is unambiguous
+add:i64_i64($0, $1)      // Full: resolves only if "add:i64_i64" is registered
+add#1($0, $1)            // Anchor: resolves anchor 1, validates base name "add"
+add:i64_i64#1($0, $1)    // Anchor + full name: exact consistency check
+count()                  // Simple: resolves if "count" base name is unambiguous
+count:()                 // Full: resolves only if "count:" is registered exactly
+```
 
 ### Aggregate Measures
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -380,11 +380,11 @@ Root[result]
 
 #### Function Name Resolution
 
-Within the plan, a function name has three parts: a `base` name (e.g. `abs`), a type signature (prefixed with a colon, e.g. `abs:i64`), and anchor (prefixed with `#`, e.g. `#4`).
+Within the plan, a function name has three parts: a `base` name (e.g. `abs`), a type signature (prefixed with a colon, e.g. `:i64`), and anchor (prefixed with `#`, e.g. `#4`).
 
-Both type signature and anchor are separably optional if the reference is unambiguous; either or both may be required to make the reference unambiguous. A function name (base, signature if present, and anchor if present) must  to exactly one function named in the `Extensions` section.
+Both type signature and anchor are separably optional if the reference is unambiguous; either or both may be required to make the reference unambiguous. A function name (base, signature if present, and anchor if present) must map to exactly one function named in the `Extensions` section.
 
-Where unambiguous, signature and anchor may both be left off (`abs($0)`), used separately (e.g. `abs:i64($0)`, `abs$4($0)`), or together for completeness (`abs:i64#4($0)`).
+Where unambiguous, signature and anchor may both be left off (`abs($0)`), used separately (e.g. `abs:i64($0)`, `abs#4($0)`), or together for completeness (`abs:i64#4($0)`).
 
 #### Examples
 
@@ -392,17 +392,17 @@ Where unambiguous, signature and anchor may both be left off (`abs($0)`), used s
 // Simple: resolves if there is exactly one function named `add`
 add($0, $1)
 // Signature: resolves only if exactly one function named `add` is registered,
-// with signature `:i64_i64`
+// with type signature `:i64_i64`
 add:i64_i64($0, $1)
 // Anchor: resolves if anchor 1 exists with base name `add`
 add#1($0, $1)            
 // Anchor + full name: resolves if anchor 1 exists with name `add` and 
-// signature `i64_i64`
+// type signature `i64_i64`
 add:i64_i64#1($0, $1)
 // Simple: resolves if there is exactly one function named `count`
 count()
 // Signature: resolves if "count:" is registered exactly once with
-// signature "" (zero arguments)
+// type signature "" (zero arguments)
 count:()
 ```
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -380,26 +380,30 @@ Root[result]
 
 #### Function Name Resolution
 
-A compound name is either *simple* (no colon, e.g. `add`) or *full* (has a colon, e.g. `add:i64_i64` or `count:`). These resolve differently:
+Within the plan, a function name has three parts: a `base` name (e.g. `abs`), a type signature (prefixed with a colon, e.g. `abs:i64`), and anchor (prefixed with `#`, e.g. `#4`).
 
-- **Simple** (`add`): matches any registered function whose base name is `add`, regardless of type signature. Succeeds only if the base name is unambiguous (exactly one registered function has that base name).
-- **Full** (`add:i64_i64`, `count:`): exact match only — the written name must match the registered name exactly.
+Both type signature and anchor are separably optional if the reference is unambiguous; either or both may be required to make the reference unambiguous. A function name (base, signature if present, and anchor if present) must  to exactly one function named in the `Extensions` section.
 
-An `anchor` explicitly identifies which registered function is meant. When present:
-- A simple name validates that the anchor's registered name shares the same base.
-- A full name validates that the anchor's registered name matches exactly.
-
-The type signature and anchor are each optional, but either or both may be required to make the reference unambiguous.
+Where unambiguous, signature and anchor may both be left off (`abs($0)`), used separately (e.g. `abs:i64($0)`, `abs$4($0)`), or together for completeness (`abs:i64#4($0)`).
 
 #### Examples
 
 ```text
-add($0, $1)              // Simple: resolves if "add" is unambiguous
-add:i64_i64($0, $1)      // Full: resolves only if "add:i64_i64" is registered
-add#1($0, $1)            // Anchor: resolves anchor 1, validates base name "add"
-add:i64_i64#1($0, $1)    // Anchor + full name: exact consistency check
-count()                  // Simple: resolves if "count" base name is unambiguous
-count:()                 // Full: resolves only if "count:" is registered exactly
+// Simple: resolves if there is exactly one function named `add`
+add($0, $1)
+// Signature: resolves only if exactly one function named `add` is registered,
+// with signature `:i64_i64`
+add:i64_i64($0, $1)
+// Anchor: resolves if anchor 1 exists with base name `add`
+add#1($0, $1)            
+// Anchor + full name: resolves if anchor 1 exists with name `add` and 
+// signature `i64_i64`
+add:i64_i64#1($0, $1)
+// Simple: resolves if there is exactly one function named `count`
+count()
+// Signature: resolves if "count:" is registered exactly once with
+// signature "" (zero arguments)
+count:()
 ```
 
 ### Aggregate Measures
--- a/src/extensions/simple.rs
+++ b/src/extensions/simple.rs
@@ -1028,7 +1028,7 @@ Type Variations:
 
     #[test]
     fn test_resolve_function_with_anchor() {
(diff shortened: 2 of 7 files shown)


wackywendell (changes requested):
Good find on the problem! However, I think we should perhaps, in fact, fix this in the opposite direction: always _include_ the colon and accept it, so the plan would look something like `Project[count:()]`. It looks slightly funny, but as that is the agreed upon syntax in both the protos and the YAML files, I think we should stick to it.

Can we do that?

Also, can we get a full round-trip plan here - showing the use of e.g. `count:` in both the simple extensions and in the plan itself?

I think that might look something like:

```
=== Extensions
URNs:
  @  1: extension:io.substrait:functions_aggregate_generic
Functions:
  # 2 @  1: count:

=== Plan
Root[num_rows]
  Project[count:() => $2]
    Read:Virtual[(1, 'alice'), (2, 'bob'), (3, 'carol') => id:i64, name:string]
```

src/parser/expression_grammar.pest
```diff
@@ -32,10 +32,13 @@ quoted_name = @{
 name = { identifier | quoted_name }
 
 // A compound name is a Substrait function compound name: base identifier followed by an optional
-// colon-delimited signature (e.g. "equal:any_any", "regexp_match_substring:str_str_i64").
+// colon-delimited signature (e.g. "equal:any_any", "count:").
+// The signature after the colon is an optional identifier: present for "equal:any_any", absent
+// for the zero-argument form "count:". Using identifier? (not identifier+) allows the trailing
+// colon with nothing after it, while keeping the same character constraints as the base name.
```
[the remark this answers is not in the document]
gord02: good catch! it closes 144.

tests/plan_roundtrip.rs
```diff
@@ -241,6 +241,97 @@ Root[a, b]
...
+Root[result]
+  Project[add($0, $1)]
+    Read[t => a:i64, b:i64]"#;
+
+    assert!(Parser::parse(plan).is_err());
+}
+
+/// `add` (no-signature) and `add:` (zero-argument) registered under the same
+/// URN.  The zero-argument form is unambiguous so it needs no anchor; the
+/// no-signature form requires an anchor because the base name `add` is shared.
```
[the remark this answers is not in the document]
gord02: This is a good point, we should probably forbid the use of `add:` function calls when there are non-zero arguments.

src/extensions/simple.rs
```diff
@@ -82,11 +82,14 @@ pub enum InsertError {
...
 ///
-/// The name before the first `:` is the *base name*; the part after is the
-/// *type-signature suffix*.  For plain names with no `:` the base name is the
-/// full name and `has_signature` returns `false`.
+/// The name before the `:` is the *base name*; the part after is the
+/// *type-signature suffix* (empty for zero-argument functions like `"count:"`).
+/// Function compound names may or may not contain a `:`:
+/// - `"add"` — no type signature
+/// - `"count:"` — conventionally used for zero-argument functions
+/// - `"equal:any_any"` — colon with a type-signature suffix
```
wackywendell: The `count:` and `equal:any_any` forms are not conceptually different: they are both of the form `<name> ":" <arguments>`; the only difference is the number of arguments

```suggestion
/// - `"count:"` — function named `count` with a zero-argument type signature suffix
/// - `"equal:any_any"` — function named `equal` with a two-argument signature suffix
```

src/extensions/simple.rs
```diff
@@ -107,12 +110,22 @@ impl CompoundName {
...
+    /// after it (e.g. `"count:"`). Distinct from no signature at all (`"count"`).
+    pub fn is_zero_arg(&self) -> bool {
+        self.index + 1 == self.name.len()
+    }
+
+    #[deprecated(
+        since = "0.6.0",
+        note = "use is_zero_arg() for the trailing-colon (zero-argument) form, \
+                or check !full().contains(':') for the no-colon form"
+    )]
```
wackywendell: Why are we deprecating this? Can we just remove it?

src/extensions/simple.rs
```diff
@@ -492,26 +505,35 @@ impl SimpleExtensions {
...
+    /// * `anchor = Some(a)` — the anchor identifies the function; the name is
+    ///   validated for consistency. An exact compound-name match is always
+    ///   accepted; no-signature and zero-argument forms also match on base name
+    ///   alone (the signature is redundant when an anchor is present).
+    /// * `anchor = None` — the name must be unambiguous on its own:
+    ///   - No signature: base-name search; fails if more than one function
+    ///     shares that base name.
+    ///   - Zero-argument: exact match first, then base-name fallback so that
+    ///     `add:()` resolves `add:i64_i64` when it is the only overload.
+    ///   - Full signature: exact match only.
```
wackywendell: Two things:

### Behavior

> `add:()` resolves `add:i64_i64` when it is the only overload.`

This should _not_ resolve like that.

`add()` should match an extension function named `add`, `add:`, or `add:i64_i64`; it didn't specify a signature, so it should match any signature.

`add:()` is quite specifically saying "the function called `add` with the type signature `()`"; it should only match the extension function  `add:`.


```suggestion
    /// * `anchor = Some(a)` — the anchor identifies the function; the name is
    ///   validated for consistency. An exact compound-name match is always
    ///   accepted; no-signature and zero-argument forms also match on base name
    ///   alone (the signature is redundant when an anchor is present).
    /// * `anchor = None` — the name must be unambiguous on its own:
    ///   - No signature: base-name search; fails if more than one function
    ///     shares that base name.
    ///   - Full signature: exact match only.
```

### Implementation

This would make more sense as two functions:

1. A function on `CompoundName`, e.g. `CompoundName::matches(&self, pattern: &str) -> bool`. The `matches` function takes a "pattern", which can either be signature-less (no `:`, such as `add`), in which case it returns true if the base name matches, regardless of signature; or it can have a signature (e.g. `count:` or `add:i64_i64`), in which case, it returns true only if its an exact match.
  - Note that just by following that, only exact-match works if the `CompoundName` is signature-less
2. This function - which then delegates the function-match part to `CompoundName::matches`

src/extensions/simple.rs
```diff
@@ -492,26 +505,35 @@ impl SimpleExtensions {
...
     pub fn resolve_function(
         &self,
         name: &str,
         anchor: Option<u32>,
     ) -> Result<ResolvedFunction<'_>, MissingReference> {
-        let resolved_anchor = match anchor {
+        let lookup = CompoundName::new(name);
+        let no_signature = !lookup.full().contains(':');
+        // True for "add" (no signature) and "add:" (zero-argument) — the caller supplied
```
wackywendell: See above - this should have the same matching logic as the other case, and if you factor that out into a `CompoundName::matches` function then that will ensure it for you.

src/extensions/simple.rs
```diff
@@ -521,13 +543,27 @@ impl SimpleExtensions {
                     ));
                 }
             }
-            None => match self.find_by_name(ExtensionKind::Function, name) {
-                Ok(a) => a,
-                Err(MissingReference::MissingName(_, _)) if !name.contains(':') => {
-                    self.find_by_base_name(ExtensionKind::Function, name)?
+            None => {
+                if no_signature {
```
wackywendell: This logic, and the logic in `find_by_base_name`, should all simply use the factored-out `CompoundName::matches` logic, and become simple loops testing matching.

src/extensions/simple.rs
```diff
@@ -107,12 +110,22 @@ impl CompoundName {
...
-    /// The full compound name, e.g. `"equal:any_any"`.
+    /// The full compound name, e.g. `"equal:any_any"` or `"count:"`.
     pub fn full(&self) -> &str {
         &self.name
     }
 
-    /// `true` when the name includes a signature suffix (contains `:`).
+    /// `true` when the name uses the zero-argument form: a colon with nothing
+    /// after it (e.g. `"count:"`). Distinct from no signature at all (`"count"`).
+    pub fn is_zero_arg(&self) -> bool {
```
wackywendell: Hm, perhaps we can drop both `is_zero_arg` and `has_type_signature`, and build this more structurally:

1. Add a new private enum:

```rust

struct FunctionNameComponents<'a> {
    // `CompoundName` with both a base name and type signature
    Full(&'a str, &'a str),
    // `CompoundName` without a type signature
    Simple(&'a str),
}
```

Then add a `CompoundName::components(&self) -> FunctionNameComponents`, and that should cover both cases.

And... I don't think this should need to be public, or even `pub(crate)`; if you have a `matches()` function, I think that should be all you need...

src/extensions/simple.rs
```diff
@@ -1015,19 +1039,35 @@ Type Variations:
...
+
+        // "like:" does not match stored "equal:any_any" → error.
+        assert!(exts.resolve_function("like:", Some(1)).is_err());
+    }
+
+    #[test]
+    fn test_resolve_function_with_anchor_zero_arg() {
+        // Zero-arg form ("name:") with explicit anchor matches by base name.
+        let exts = make_resolve_extensions();
+        assert_eq!(exts.resolve_function("equal:", Some(1)).unwrap().anchor, 1);
```
wackywendell: `equal` should match; `equal:` should _not_match.

tests/plan_roundtrip.rs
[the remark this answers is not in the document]
wackywendell: This is more general: we should either check argument number / type matches generally (zero args or more), or not - e.g. calling `add:i64_i64()` is just as incorrect as `add:($0, $1)`. Right now, we don't check argument matches at all; I'm not sure we need to, but we could make an issue for it.

wackywendell (changes requested):
Two main comments:

1. The documentation and logic should not treat "zero arg" as a special, separate, third case. It is simply the "base" case of the type-signature form - grammatically and logically it should be treated the same as the multi-arg form. It's like how you might have `null`, `[]`, `[1]`, `[2, 3]`, ...; the "empty list" is just the base case for a list of multiple things, although `null` is special.
2. Related: we should update the `GRAMMAR.md` file to describe how function matching works.
3. Implementation-wise, we should factor out the matching logic from the "find" logic; it will both simplify the code and reduce the possibility of inconsistencies. (e.g. whether `name#anchor` matches `full_name` should be the same logic as whether `name` matches `full_name`).

src/extensions/simple.rs
gord02: If the matches() function takes a string pattern as input we don't need the enum. We could change matches() to take CompoundName — which would make the enum useful and keep the logic expressed as a structural match, but that might be more of a redesign that it is worth?

wackywendell: Good point! `matches` is now elegantly simple and covers all these use cases, so let's keep it to that!

wackywendell:
Code looks good! Last change needed is the `GRAMMAR.md` update to clarify the allowed syntax for function calls - namely, that both or either type signature and anchor are optional insofar as the reference is unambiguous; either or both may be required to disambiguate.

GRAMMAR.md
```diff
@@ -366,16 +366,42 @@ Root[result]
...
+A compound name is either *simple* (no colon, e.g. `add`) or *full* (has a colon, e.g. `add:i64_i64` or `count:`). These resolve differently:
+
+- **Simple** (`add`): matches any registered function whose base name is `add`, regardless of type signature. Succeeds only if the base name is unambiguous (exactly one registered function has that base name).
+- **Full** (`add:i64_i64`, `count:`): exact match only — the written name must match the registered name exactly.
+
+An `anchor` explicitly identifies which registered function is meant. When present:
+- A simple name validates that the anchor's registered name shares the same base.
+- A full name validates that the anchor's registered name matches exactly.
+
+The type signature and anchor are each optional, but either or both may be required to make the reference unambiguous.
```
wackywendell: The implementation separates 'full' and 'simple', and type signature and anchor, and treats one as part of the lookup and one as part of the validation, but from the _user_ perspective, it's simpler: the base is required; the signature and anchor are optional; but the lookup will match whatever is provided. For the user, there is no difference between "failed to validate" and "didn't match" - both return an error.

```suggestion
Within the plan, a function name has three parts: a `base` name (e.g. `abs`), a type signature (prefixed with a colon, e.g. `abs:i64`), and anchor (prefixed with `#`, e.g. `#4`).

Both type signature and anchor are separably optional if the reference is unambiguous; either or both may be required to make the reference unambiguous. A function name (base, signature if present, and anchor if present) must  to exactly one function named in the `Extensions` section.

Where unambiguous, signature and anchor may both be left off (`abs($0)`), used separately (e.g. `abs:i64($0)`, `abs$4($0)`), or together for completeness (`abs:i64#4($0)`).
```

GRAMMAR.md
```diff
@@ -366,16 +366,42 @@ Root[result]
...
+
+#### Examples
+
+```text
+add($0, $1)              // Simple: resolves if "add" is unambiguous
+add:i64_i64($0, $1)      // Full: resolves only if "add:i64_i64" is registered
+add#1($0, $1)            // Anchor: resolves anchor 1, validates base name "add"
+add:i64_i64#1($0, $1)    // Anchor + full name: exact consistency check
+count()                  // Simple: resolves if "count" base name is unambiguous
+count:()                 // Full: resolves only if "count:" is registered exactly
```
wackywendell: As above - let's put signature and anchor in similar terms here, and describe the behavior as matching exactly once. Also, instead of using "full" to mean "base name plus signature", I think the word "signature" is clearer, and use "full" to mean base + signature + anchor.

```suggestion
// Simple: resolves if there is exactly one function named `add`
add($0, $1)
// Signature: resolves only if exactly one function named `add` is registered,
// with signature `:i64_i64`
add:i64_i64($0, $1)
// Anchor: resolves if anchor 1 exists with base name `add`
add#1($0, $1)            
// Anchor + full name: resolves if anchor 1 exists with name `add` and 
// signature `i64_i64`
add:i64_i64#1($0, $1)
// Simple: resolves if there is exactly one function named `count`
count()
// Signature: resolves if "count:" is registered exactly once with
// signature "" (zero arguments)
count:()
```

wackywendell (approved):
One final comment on the Grammar, to put it more in terms of user-level concepts, rather than terms that are more implementation-based.

PR #153, opened 2026-06-11
fix: require output type on scalar and aggregate function calls

### LINKED ISSUE #145: [BUG] Scalar Function Expressions Output Type
## Description

The parsing rule for scalar function expressions(function_call) has type as optional. This is because the function can be displayed without the output type if there is no overlapping function names. 

For downstream consumers of the generated proto such as substrait-java, output_type is required. 

## Steps to Reproduce

The following substrait text parses to proto without setting the output type. 

```
=== Extensions
URNs:
  @  1: extension:io.substrait:functions_arithmetic
Functions:
  #  1 @  1: add:i64_i64

=== Plan
Root["1 + 1"]
  Project[add(1, 1)]
    Read:Virtual[_ => ]
```

The produced proto is as follows missing the output_type and throws an exception `Type is not set: `. 
```
 scalar_function {
            function_reference: 1
            arguments {
              value {
                literal {
                  i64: 1
                }
              }
            }
            arguments {
              value {
                literal {
                  i64: 1
                }
              }
            }
          }
```
## Expected Behavior 
The proto should be parsed into substrait-java POJO. Instead the downstream consumers reject the generated plan.

gord02:
## Description

The grammar defined the output type annotation on function_call as optional ((":" type)?), allowing plans like add(1, 1) to be written and parsed without specifying a return type. This produced a ScalarFunction proto with output_type unset, which goes against the substrait semantics and breaks in downstream consumers such as substrait-java, which rejects with `Type is not set"`.

Additionally, the textifier gated output type rendering on the fn_types option (default: off), meaning round-tripped plans would silently drop the type even when it was present in the proto.

### Changes
Grammar & parser (expression_grammar.pest, expressions.rs)
- Made the output type required on function_call: ":" type (no trailing ?)
- Updated the parser to use pop instead of try_pop for the type rule, so output_type is always Some

Textifier (textify/expressions.rs)
- ScalarFunction and AggregateFunction now always render the output type, no longer gated on fn_types

Cleanup
- Removed the show_expression_types CLI flag (--show-expression-types) and the fn_types field from OutputOptions, both of which are now obsolete
- Updated doc comments in foundation.rs, types.rs, and GRAMMAR.md to reflect that output types are required

Tests & examples
- Updated unit tests, integration tests, doc tests, and example plans to include explicit output types on every function call

## Type of Change
- [x] Bug fix

## Testing
- [x] All existing tests pass

## Related Issues
Closes #145

### DIFF
--- a/API.md
+++ b/API.md
@@ -33,7 +33,7 @@ Functions:
   ## 10 @  1: add
 
 === Plan
-Project[$0, $1, add($0, $1)]
+Project[$0, $1, add($0, $1):i32?]
   Read[table1 => col1:i32?, col2:i32?]
 "#;
 
@@ -240,8 +240,8 @@ Functions:
   # 12 @  2: count
 === Plan
 Root[result]
-  Aggregate[$0 => $0, sum($1), count($1)]
-    Project[$0, add($1, $2)]
+  Aggregate[$0 => $0, sum($1):i32?, count($1):i64]
+    Project[$0, add($1, $2):i32?]
       Read[table1 => category:string, col1:i32?, col2:i32?]
 ```
 
@@ -258,7 +258,7 @@ Each relation is displayed on a single line with the format:
 
 - **Field references**: `$0`, `$1`, etc.
 - **Literals**: `42`, `'hello'`, `true`
-- **Function calls**: `add($0, $1)`, `sum($2)` (scalar and aggregate functions)
+- **Function calls**: `add($0, $1):i64`, `sum($2):i64`
 - **Types**: `i32`, `string?`, `list<i64>`
 
 ## Configuration Options
--- a/API.md
+++ b/API.md
@@ -338,7 +338,6 @@ cat plan.substrait | substrait-explain convert -f text -t json > plan.json
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
 - `--show-literal-types` - Show type annotations on literals
-- `--show-expression-types` - Show type annotations on expressions
 - `--verbose` - Show detailed progress information
 
 #### Validate Command
@@ -370,7 +369,7 @@ substrait-explain validate -i example-plans/basic.substrait
 substrait-explain validate -i example-plans/simple.substrait
 
 # Convert with verbose output and type information
-substrait-explain convert -f text -t json --show-literal-types --show-expression-types --verbose -i example-plans/basic.substrait
+substrait-explain convert -f text -t json --show-literal-types --verbose -i example-plans/basic.substrait
 
 # Roundtrip test: text → protobuf → text
 substrait-explain convert -f text -t protobuf -i plan.substrait -o plan.pb
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -76,7 +76,7 @@ Functions:
 
 === Plan
 Root[result]
-  Project[$0, $1, add($0, $1)]
+  Project[$0, $1, add($0, $1):i64]
     Read[orders => quantity:i32?, price:i64]
 # "#;
 #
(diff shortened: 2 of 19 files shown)


src/textify/foundation.rs
```diff
@@ -510,6 +506,7 @@ pub(crate) trait Scope: Sized {
         }
     }
 
+    #[allow(dead_code)]
```
gord02, line 510: This might be useful in the future but was only used for the function output type. Should we leave it as is here or remove it entirely?

Justfile
```diff
@@ -20,6 +20,9 @@ fmt:
 # Run the tests.
 test:
     cargo test --all-features
+    cargo run --example basic_usage --features protoc
```
gord02, line 23: Just test wasn't running everything the CI was, particularly, the files in examples/.

src/textify/expressions.rs
```diff
@@ -417,7 +417,7 @@ impl Textify for ScalarFunction {
         };
 
         let output = OutputType(self.output_type.as_ref());
-        let output_type = ctx.optional(&output, ctx.options().fn_types);
+        let output_type = ctx.display(&output);
```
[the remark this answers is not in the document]
gord02: This is resolved.  The branch changed OutputType to treat None as an error: it now calls ctx.failure(PlanError::invalid(...)), which emits an error token (e.g. !{OutputType}) and records a FormatError. The textifier unconditionally calls ctx.display(&output) rather than gating on visibility, so there is no silent omission path.

src/textify/types.rs
```diff
@@ -212,10 +212,7 @@ impl<'a> Textify for NamedAnchor<'a> {
     }
 }
 
-/// The type desciptor of the output of a function call.
-///
-/// This is optional, and if present, it must be the last argument in the
-/// function call.
+/// The output type of a function call, rendered as `:type` after the argument list.
 #[derive(Debug, Copy, Clone)]
 pub struct OutputType<T: Deref<Target = proto::Type>>(pub Option<T>);
```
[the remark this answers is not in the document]
gord02: This is resolved. We emit an error token rather than silently omitting the type.

src/parser/expression_grammar.pest
```diff
@@ -146,10 +146,10 @@ argument_list = { "(" ~ (expression ~ (sp ~ "," ~ sp ~ expression)*)? ~ ")" }
 // - Optional Anchor, e.g. #1
 // - Optional URN Anchor, e.g. @1
 // - Arguments `()` are required, e.g. (1, 2, 3)
-// - Optional output type annotation after closing paren, e.g. :i64
+// - Required output type annotation after closing paren, e.g. :i64
```
wackywendell, line 149: This should be updated in `GRAMMAR.md` as well

src/textify/foundation.rs
wackywendell: Hm, this seems like it should be useful; I'm surprised that was the only place using it. I'd suggest leaving it and adding a comment
```suggestion
    // TODO: this was last used on 2026-06-15, before a refactor removed it; if its not used again, perhaps we should drop it?
    #[allow(dead_code)]
```

src/cli.rs
```diff
@@ -148,21 +144,13 @@ impl Cli {
...
-        show_expression_types: bool,
-    ) -> OutputOptions {
+    fn create_output_options(&self, show_literal_types: bool) -> OutputOptions {
         let mut options = OutputOptions::default();
 
         if show_literal_types {
             options.literal_types = Visibility::Always;
         }
 
-        if show_expression_types {
```
wackywendell, line 162: Are functions the only kind of expression with an output type? Do they exist on any other kinds of expressions?

GRAMMAR.md
```suggestion
`function_call := compound_name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
```

Output type is no longer optional. Also update down below

API.md
wackywendell: This `show-expression-types` option was deleted, right?

wackywendell (changes requested):
The basic change looks fine, but there's a couple issues:

1. You need to update `GRAMMAR.md`
2. `API.md` documents `--show-expression-types` - if you're removing it, you should remove it from the docs
3. More tests need to be updated and fixed; they're failing CI because they don't have type ascriptions.

src/cli.rs
gord02: Output_type is a field only on functions only.

wackywendell (approved):
Great, thanks for doing this!

PR #154, opened 2026-06-12
feat: u!-prefix support for user-defined types and function signatures

### LINKED ISSUE #148: [BUG] u!TypeName argument notation in function compound names
## Description

[substrait semantics](https://substrait.io/extensions/#referencing-user-defined-types) dictates that user defined types are referenced with the u! prefix followed by the type name.

Substrait explain text in the form of `json_extract_path:u!json_str` cannot be pared into proto since `!` isn't apart of the character set for identifiers. 

## Proposed Solution
We need to account for this case of `u!` in the compound name. A compound_name should be a combination of identifiers and `u!` prefixed identifiers. 
```
sig_segment   = @{ "u!" ~ identifier | identifier }
compound_name = @{ identifier ~ (":" ~ sig_segment ~ ("_" ~ sig_segment)*)? }
```

This:
- Handles json_extract_path:u!json_str — sig_segment matches u!json_str 
- Stays backward-compatible with equal:any_any — sig_segment greedily matches any_any as one identifier 
- Stays backward-compatible with regexp_match_substring:str_str_i64

gord02:
User-defined types with u! prefix support

The text format now fully supports user-defined types with the optional u! prefix used in Substrait YAML files. Both u!json and json refer to the same registered type — the prefix is normalized at storage time so declarations and plan references can use either form interchangeably. The canonical output always uses the bare name.

The u! prefix is accepted on type declarations in the Extensions section but is rejected on function and type-variation declarations, matching the Substrait spec. Function signatures may include u!-prefixed argument type codes (e.g. json_extract_path:u!json_str).

## Type of Change
- [x] Bug fix
- [x] New feature

## Testing
- [x] All existing tests pass

## Related Issues

Closes #148

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -210,13 +210,13 @@ The type syntax in this grammar follows the [standard Substrait type definition 
 All types follow this general pattern:
 
 ```text
-type := "u!"? name anchor? urn_anchor? nullability? parameters?
+type := type_name anchor? urn_anchor? nullability? parameters?
+type_name := "u!" identifier | identifier
 ```
 
 Where:
 
-- **`"u!"`** - Optional prefix for user-defined types
-- **`name`** - Type name (case-insensitive, lowercase preferred)
+- **`type_name`** - Type name(case-insensitive, lowercase preferred), plain (e.g. `i64`, `geo_point`) or with `u!` prefix for user-defined types (e.g. `u!json`). The prefix is part of the name and must match the declaration exactly.
 - **`anchor`**` := "#" integer` - Extension anchor (e.g., `#10`)
 - **`urn_anchor`**` := "@" integer` - URN anchor (e.g., `@1`)
 - **`nullability`**` := "?"` - Optional nullability indicator (defaults to non-nullable)
@@ -291,13 +291,14 @@ User-defined types extend the standard Substrait UDT syntax to support anchors a
 
 #### Syntax
 
-`"u!"? name anchor? urn_anchor? nullability? parameters?`
+`type_name anchor? urn_anchor? nullability? parameters?`
+
+where `type_name` is `"u!" identifier | identifier`.
 
 #### Key differences from standard Substrait
 
-- The `u!` prefix is optional (can be omitted when anchors are present)
 - Adds optional `anchor` and `urn_anchor` for extension references
-- Maintains compatibility with standard Substrait UDT syntax
+- The `u!` prefix is part of the type name: a type declared as `u!json` must be referenced as `u!json`, not bare `json`
 
 #### Examples
 
@@ -311,14 +312,14 @@ URNs:
   @  2: https://example.com/functions
 Types:
   ##  8 @  1: point
-  ##  9 @  1: custom_type
+  ##  9 @  1: u!custom_type
 Functions:
   ## 10 @  2: add
 
 === Plan
 Root[result]
   Project[$0, $1, $2]
-    Read[data => point_field:point#8@1?<i8>, custom_field:custom_type#9, prefixed_field:u!custom_type]
+    Read[data => point_field:point#8@1?<i8>, custom_field:u!custom_type#9, prefixed_field:u!custom_type]
 # "#;
 #
 # let plan = Parser::parse(plan_text).unwrap();
@@ -366,11 +367,24 @@ Root[result]
 
 #### Syntax
 
-`function_call := name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
+`function_call := compound_name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
+
+where `compound_name` is the base function name with an optional colon-delimited type-signature suffix:
+
+```text
+compound_name := type_name (":" function_args?)?
+function_args := type_arg ("_" type_arg)*
+type_arg      := "u!" type_code | type_code
+type_code     := ASCII_ALPHA ASCII_ALPHANUMERIC*
+```
+
+Examples of compound names: `add`, `equal:any_any`, `count:` (empty signature), `json_extract_path:u!json_str`, `bar:u!arg_u!arg`.
+
+The `u!` prefix may appear in signature argument slots as well as in the base name. Note that `type_code` (used in signatures) does not allow underscores — `_` is the argument separator.
 
 #### Components
 
-- `name` - function name
+- `compound_name` - function name with optional signature suffix (see above)
 - `anchor` - optional anchor (e.g., `#10`)
 - `urn_anchor` - optional URN anchor (e.g., `@1`)
 - `expression` - as above
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -218,7 +218,7 @@ type := ("u!")? identifier anchor? urn_anchor? nullability? parameters?
 
 Where:
 
-- **`identifier`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
+- **`identifier`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `my_type`. `u!my_type` syntax is accepted, but not recommended; the `u!` will be dropped - e.g. both `u!json` and `json` refer to the same extension.
 - **`anchor`**` := "#" integer` - Extension anchor (e.g., `#10`)
 - **`urn_anchor`**` := "@" integer` - URN anchor (e.g., `@1`)
 - **`nullability`**` := "?"` - Optional nullability indicator (defaults to non-nullable)
@@ -368,7 +368,7 @@ Root[result]
 
 #### Syntax
 
-`function_call := compound_function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
+`function_call := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
 
 where `function_signature` is the function base name with an optional colon-delimited type-signature suffix:
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -210,13 +210,12 @@ The type syntax in this grammar follows the [standard Substrait type definition 
 All types follow this general pattern:
 
 ```text
-type := type_name anchor? urn_anchor? nullability? parameters?
-type_name := "u!" identifier | identifier
+user_defined_type := ("u!")? (identifier | quoted_name) anchor? urn_anchor? nullability? parameters?
 ```
 
 Where:
 
-- **`type_name`** - Type name(case-insensitive, lowercase preferred), plain (e.g. `i64`, `geo_point`) or with `u!` prefix for user-defined types (e.g. `u!json`). The prefix is part of the name and must match the declaration exactly.
+- **`identifier | quoted_name`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
 - **`anchor`**` := "#" integer` - Extension anchor (e.g., `#10`)
 - **`urn_anchor`**` := "@" integer` - URN anchor (e.g., `@1`)
 - **`nullability`**` := "?"` - Optional nullability indicator (defaults to non-nullable)
@@ -291,14 +290,13 @@ User-defined types extend the standard Substrait UDT syntax to support anchors a
 
 #### Syntax
 
-`type_name anchor? urn_anchor? nullability? parameters?`
-
-where `type_name` is `"u!" identifier | identifier`.
+`("u!")? (identifier | quoted_name) anchor? urn_anchor? nullability? parameters?`
 
 #### Key differences from standard Substrait
 
 - Adds optional `anchor` and `urn_anchor` for extension references
-- The `u!` prefix is part of the type name: a type declared as `u!json` must be referenced as `u!json`, not bare `json`
+- The `u!` prefix is accepted in type declarations but normalized at storage time, so `u!json` and `json` in a declaration refer to the same type. It is an error to use `u!` on function or type-variation declarations.
+- In plan references both `u!json` and `json` are accepted and resolve to the same anchor. The canonical output always uses the bare name.
 
 #### Examples
 
@@ -312,14 +310,14 @@ URNs:
   @  2: https://example.com/functions
 Types:
   ##  8 @  1: point
-  ##  9 @  1: u!custom_type
+  ##  9 @  1: custom_type
 Functions:
   ## 10 @  2: add
 
 === Plan
 Root[result]
   Project[$0, $1, $2]
-    Read[data => point_field:point#8@1?<i8>, custom_field:u!custom_type#9, prefixed_field:u!custom_type]
+    Read[data => point_field:point#8@1?<i8>, custom_field:custom_type#9, prefixed_field:u!custom_type]
 # "#;
 #
 # let plan = Parser::parse(plan_text).unwrap();
@@ -367,24 +365,22 @@ Root[result]
 
 #### Syntax
 
-`function_call := compound_name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
+`function_call := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" (":" type)?`
 
-where `compound_name` is the base function name with an optional colon-delimited type-signature suffix:
+where `function_signature` is the function base name with an optional colon-delimited type-signature suffix:
 
 ```text
-compound_name := type_name (":" function_args?)?
-function_args := type_arg ("_" type_arg)*
-type_arg      := "u!" type_code | type_code
-type_code     := ASCII_ALPHA ASCII_ALPHANUMERIC*
+function_signature := identifier (":" argument_signature?)?
+argument_signature := short_arg_type ("_" short_arg_type)*
+short_arg_type     := "u!" short_arg_name | short_arg_name
+short_arg_name     := ASCII_ALPHA ASCII_ALPHANUMERIC*
 ```
 
-Examples of compound names: `add`, `equal:any_any`, `count:` (empty signature), `json_extract_path:u!json_str`, `bar:u!arg_u!arg`.
-
-The `u!` prefix may appear in signature argument slots as well as in the base name. Note that `type_code` (used in signatures) does not allow underscores — `_` is the argument separator.
+Examples of function signatures: `add`, `equal:any_any`, `count:` (empty signature), `json_extract_path:u!json_str`, `bar:u!arg_u!arg`.
 
 #### Components
 
-- `compound_name` - function name with optional signature suffix (see above)
+- `function_signature` - function name with optional signature suffix (see above)
 - `anchor` - optional anchor (e.g., `#10`)
 - `urn_anchor` - optional URN anchor (e.g., `@1`)
 - `expression` - as above
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -213,12 +213,12 @@ The type syntax in this grammar follows the [standard Substrait type definition 
 All types follow this general pattern:
 
 ```text
-type := ("u!")? (identifier | quoted_name) anchor? urn_anchor? nullability? parameters?
+type := ("u!")? identifier anchor? urn_anchor? nullability? parameters?
 ```
 
 Where:
 
-- **`identifier | quoted_name`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
+- **`identifier`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
 - **`anchor`**` := "#" integer` - Extension anchor (e.g., `#10`)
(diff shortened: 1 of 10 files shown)


src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
 name = { identifier | quoted_name }
 
-// A compound name is a Substrait function compound name: base identifier followed by an optional
-// colon-delimited signature (e.g. "equal:any_any", "regexp_match_substring:str_str_i64").
-// The rule is atomic so no whitespace is allowed around the colon, making it unambiguous
-// with the output-type annotation which always follows the closing parenthesis.
-compound_name = @{ identifier ~ (":" ~ identifier)? }
+// A plain or u!-prefixed type name, used in schema, cast positions, and function base names.
+// Examples: "geo_point", "u!geo_point"
+type_name = @{ "u!" ~ identifier | identifier }
```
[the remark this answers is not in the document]
gord02: type_name and udt_name are two different things. a type name _can_ be used in schema, cast and function names.

GRAMMAR.md
```diff
@@ -210,13 +210,13 @@ The type syntax in this grammar follows the [standard Substrait type definition
 All types follow this general pattern:
 
 ```text
-type := "u!"? name anchor? urn_anchor? nullability? parameters?
+type := type_name anchor? urn_anchor? nullability? parameters?
+type_name := "u!" identifier | identifier
 ```
```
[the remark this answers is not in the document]
gord02: We may not need all types to support quoted names, therefore I think this can be ignored.

src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
+
+// Examples: 'geo_point', 'u!geo_point', 'u!"my type"'
+udt_name = @{ "u!" ~ (identifier | quoted_name) | identifier | quoted_name }
```
wackywendell: Why is there `udt_name` separate from `type_name`? Why isn't a `type_name` either `udt_name` or `identifier`?

Thinking about this, and looking at the spec, I believe:

- All simple extensions (type names, function names, etc.) should be named with an `identifier`, so there's no need for `quoted_name` here
- `u!` is permitted for UDTs, and included in argument signatures, but is not part of function names

So I think you can drop both `udt_name` and `type_name`, and use `"u!"? ~ identifier` in `user_defined_type` as before. If you want, you can keep `udt_name = @{ "u!"? ~ identifier }`, but I'd put it right above `user_defined_type`, as that's the only place its used

src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
+// argument type in a function signature: plain or u!-prefixed.
+// Examples: "i64", "u!json"
+type_arg = @{ "u!" ~ type_code | type_code }
+
+// suffix of a compound name: underscore-separated type_args.
+function_args = @{ type_arg ~ ("_" ~ type_arg)* }
+
+// A Substrait compound name: base name with an optional colon-delimited signature suffix.
+// Examples: "add", "equal:any_any", "count:", "json_extract_path:u!json_str", "bar:u!arg_u!arg".
+compound_name = @{ type_name ~ (":" ~ function_args?)? }
```
wackywendell: Two things:

1. Let's use the same naming as the official docs
2. It is confusing to use something called `type_name` here, and also incorrect; I believe `u!` is not allowed here, and is part of `type_name`. But as its not a type, its a function, `type_name` seems wrong anyway.

```suggestion
function_signature = @{ identifier ~ (":" ~ argument_signature?)? }
```

src/parser/expression_grammar.pest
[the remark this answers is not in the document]
wackywendell: I'm confused by this too; `type_name` seems more narrow than `udt_name`, which is confusing, and then we use `type_name` for the function name in a compound name? That doesn't seem right...

src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
+// No underscores since "_" is the argument separator.
+// Examples: "i64", "str", "any", "json"
+type_code = @{ ASCII_ALPHA ~ ASCII_ALPHANUMERIC* }
+
+// argument type in a function signature: plain or u!-prefixed.
+// Examples: "i64", "u!json"
+type_arg = @{ "u!" ~ type_code | type_code }
+
+// suffix of a compound name: underscore-separated type_args.
+function_args = @{ type_arg ~ ("_" ~ type_arg)* }
```
wackywendell: Let's use the same naming as [the official Substrait docs](https://substrait.io/extensions/), for clarity

```suggestion
// The type signature, used as the suffix of a function name: underscore-separated type_args
argument_signature = @{ short_arg_type ~ ("_" ~ short_arg_type)* }
```

src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
+// Examples: 'geo_point', 'u!geo_point', 'u!"my type"'
+udt_name = @{ "u!" ~ (identifier | quoted_name) | identifier | quoted_name }
+
+// No underscores since "_" is the argument separator.
+// Examples: "i64", "str", "any", "json"
+type_code = @{ ASCII_ALPHA ~ ASCII_ALPHANUMERIC* }
+
+// argument type in a function signature: plain or u!-prefixed.
+// Examples: "i64", "u!json"
+type_arg = @{ "u!" ~ type_code | type_code }
```
wackywendell: Let's use the same naming as [the official Substrait docs](https://substrait.io/extensions/), for clarity

```suggestion
short_arg_type = @{ "u!" ~ short_arg_name | short_arg_name }
```

src/parser/expression_grammar.pest
```diff
@@ -31,11 +31,27 @@ quoted_name = @{
...
+// A plain or u!-prefixed type name, used in schema, cast positions, and function base names.
+// Examples: "geo_point", "u!geo_point"
+type_name = @{ "u!" ~ identifier | identifier }
+
+// Examples: 'geo_point', 'u!geo_point', 'u!"my type"'
+udt_name = @{ "u!" ~ (identifier | quoted_name) | identifier | quoted_name }
+
+// No underscores since "_" is the argument separator.
+// Examples: "i64", "str", "any", "json"
+type_code = @{ ASCII_ALPHA ~ ASCII_ALPHANUMERIC* }
```
wackywendell: Let's use the same naming as [the official Substrait docs](https://substrait.io/extensions/), for clarity

```suggestion
short_arg_name = @{ ASCII_ALPHA ~ ASCII_ALPHANUMERIC* }
```

src/parser/expressions.rs
```diff
@@ -1100,9 +1100,23 @@ mod tests {
         assert_eq!(parse_compound_name("add:i64_i64").full(), "add:i64_i64");
     }
 
+    #[test]
+    fn test_compound_name_user_defined_type_base() {
+        // u! prefix on the base name, used in extension type declarations e.g. "# 5 @ 1: u!json"
+        // u!json name, base() and full() are identical by design since there's no :suffix.
+        assert_eq!(parse_compound_name("u!json").full(), "u!json");
```
wackywendell: This makes no sense. A "compound name" with format `a:b` is only used for functions, not types; `u!json` should not be a compound name.

src/parser/expressions.rs
```diff
@@ -1100,9 +1100,23 @@ mod tests {
...
+    #[test]
+    fn test_compound_name_trailing_colon_grammar() {
+        // "count:" (trailing colon, zero-arg type signature) must parse fully at grammar level.
+        let pairs = ExpressionParser::parse(Rule::compound_name, "count:").unwrap();
+        assert_eq!(pairs.as_str(), "count:");
+    }
+
     #[test]
     fn test_compound_name_stops_at_opening_paren() {
-        // where is the paren ???
```
wackywendell, line 1135: 👍  That is a weird comment

src/parser/types.rs
```diff
@@ -476,4 +476,91 @@ mod tests {
...
+            .unwrap();
+        assert!(
+            parse_user_defined_type(&extensions, pair).is_err(),
+            "u!json should not match an extension declared as json"
+        );
+    }
+
+    #[test]
+    fn test_udt_prefix_mismatch_plain_against_u_prefix_registration() {
+        // Extension registered with "u!" but plan uses bare "json" — must fail.
```
wackywendell: Actually, I think it would be useful if this both succeeded and was the default option for output. I don't see a need to _require_ `u!` everywhere like that; in something intended to be human-readable, it's just noise.

My suggestion: either drop it from output, or make it an `OutputOptions` option to include the `u!`, off by default.

src/parser/types.rs
```diff
@@ -476,4 +476,91 @@ mod tests {
...
+                    nullability: Nullability::Nullable as i32,
+                    type_parameters: vec![],
+                }))
+            }
+        );
+    }
+
+    #[test]
+    fn test_udt_prefix_mismatch_u_prefix_against_plain_registration() {
+        // Extension registered without "u!" but plan uses "u!json" — must fail.
```
wackywendell: Hm, perhaps this should in fact succeed? The `u!` is not adding any information (if its in the extensions section, it is by definition user-defined), and just adding noise; it will be more human-readable without.

src/textify/expressions.rs
```diff
@@ -1099,4 +1099,25 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_cast_to_user_defined_type_textifies_with_u_prefix() {
+        // Cast to a u!-prefixed UDT must emit "::u!json", not "::json".
```
wackywendell: As above - I think by default it should be without the `u!`.

GRAMMAR.md
```diff
@@ -291,13 +291,13 @@ User-defined types extend the standard Substrait UDT syntax to support anchors a
...
-`"u!"? name anchor? urn_anchor? nullability? parameters?`
+`udt_name anchor? urn_anchor? nullability? parameters?`
 
+where `udt_name` is `"u!" (identifier | quoted_name) | identifier | quoted_name`.
 #### Key differences from standard Substrait
 
-- The `u!` prefix is optional (can be omitted when anchors are present)
 - Adds optional `anchor` and `urn_anchor` for extension references
-- Maintains compatibility with standard Substrait UDT syntax
+- The `u!` prefix is part of the type name: a type declared as `u!json` must be referenced as `u!json`, not bare `json`
```
wackywendell: Let's reverse this, for ease of use:

```suggestion
- The `u!` prefix is not part of the type name: a type declared as `u!json` or as `json` can be referenced as `u!json` or as simply `json`.
```

wackywendell (changes requested):
So I think we should clarify the grammar here:

1. From what I can tell from the docs, the only place we should expect to find a `u!` in a protobuf file is in an argument signature. It should not be in the type name, nor in the function name. Given that:
  1. I think we should allow `u!` in function signatures, because its expected there, and match based on it
  2. however, we should **not** allow it in function _names_
  3. And for type names in both the extensions section and in plans (schemas, casts, etc.), we should _allow_ but not _require_ it
  4. And also, I think it would be nice to match type names regardless of `u!` - so `u!json` in the `Extensions` section can be matched by `col:json` below, and vice versa

src/extensions/simple.rs
```diff
@@ -220,6 +223,16 @@ impl SimpleExtensions {
...
+        // Normalize type names: strip the optional "u!" prefix so that "u!json" and "json"
+        // are stored identically and resolve to the same anchor.
+        // For other extension kinds, "u!" prefix is invalid per the Substrait spec.
+        let name = if kind == ExtensionKind::Type {
+            name.strip_prefix("u!").map(str::to_string).unwrap_or(name)
+        } else if name.starts_with("u!") {
+            return Err(InsertError::InvalidName { kind, name });
+        } else {
+            name
+        };
```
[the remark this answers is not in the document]
gord02: The PR description is out of date.

src/parser/expression_grammar.pest
```diff
@@ -195,12 +198,14 @@ urn = @{ (!whitespace ~ ANY)+ }
...
+// Format: #anchor@urn_ref: simple_extension_name
+// Examples: #10@1: my_func   #1@2: equal:any_any   #5@1: json   #11@1: u!point
 // `anchor` is defined above as "#" ~ integer.
 // `urn_anchor` is defined above as "@" ~ integer.
-// `compound_name` accepts both plain names and compound names with signatures.
-simple_extension = { anchor ~ sp ~ urn_anchor ~ sp ~ ":" ~ sp ~ compound_name }
+// The u! prefix is accepted at grammar level for all extension kinds, but is only valid
+// semantically for Type declarations; Function and TypeVariation names must be bare.
+simple_extension_name = @{ "u!"? ~ identifier ~ (":" ~ argument_signature?)? }
+simple_extension = { anchor ~ sp ~ urn_anchor ~ sp ~ ":" ~ sp ~ simple_extension_name }
```
wackywendell, line 207: The comment here is fair, but let's validate / enforce that: `u!` should be allowed only on type names; function signatures only on function names.

This is partly pre-existing, so I do not think we need a big refactor here (e.g. multiple rules, multiple `ParsePair`-implementing types). But could we add a small section-aware validation where the declaration is parsed?

src/parser/expression_grammar.pest
```diff
@@ -120,19 +130,12 @@ precision_time_type         = { ^"precisiontime"        ~ nullability ~ "<" ~ in
...
-// - If the type is nullable, the nullability is required.
-// - The parameters are required if they exist for this type.
+// A user-defined type expression. The u! prefix is optional and ignored for lookup purposes;
+// both "json" and "u!json" resolve to the same extension.
+// Examples: "point#8@2?<i8>", "json?", "u!json?", "geo_point#5@1", "u!geo_point#5@1"
+// Anchor is required when the name is ambiguous across URNs; URN anchor is optional.
 user_defined_type = {
-    "u!"? ~ name ~ anchor? ~ urn_anchor? ~ nullability ~ parameters?
+    "u!"? ~ (identifier | quoted_name) ~ anchor? ~ urn_anchor? ~ nullability ~ parameters?
 }
```
wackywendell, line 140: Let's make the behavior here consistent: either allow quoted names for types and functions, references and extensions, or disallow.

We're pretty close to disallowing (I think this is the only place that allows) and support for quoted names would be complex, so... let's just go ahead and disallow; a clear error is better than confused partial support. Since full quoted-name support is separate from the `u!` change, can we make this identifier-only for now and leave a TODO?

```suggestion
// TODO: Do we need to support quoted names / special characters in type/function names?
user_defined_type = {
    "u!"? ~ identifier ~ anchor? ~ urn_anchor? ~ nullability ~ parameters?
}
```

src/extensions/simple.rs
```diff
@@ -228,6 +236,16 @@ impl SimpleExtensions {
         anchor: u32,
         name: String,
     ) -> Result<(), InsertError> {
+        // Normalize type names: strip the optional "u!" prefix so that "u!json" and "json"
+        // are stored identically and resolve to the same anchor.
+        // For other extension kinds, "u!" prefix is invalid per the Substrait spec.
+        let name = if kind == ExtensionKind::Type {
+            name.strip_prefix("u!").map(str::to_string).unwrap_or(name)
+        } else if name.starts_with("u!") {
+            return Err(InsertError::InvalidName { kind, name });
```
wackywendell, line 245: This is reasonable as a final guard, but we should probably enforce the `u!` limits (types only) and function-signature-only at the parser, so that the user gets that feedback early, linked to a line/column number.

GRAMMAR.md
```diff
@@ -213,13 +213,12 @@ The type syntax in this grammar follows the [standard Substrait type definition
 All types follow this general pattern:
 
 ```text
-type := "u!"? name anchor? urn_anchor? nullability? parameters?
+type := ("u!")? (identifier | quoted_name) anchor? urn_anchor? nullability? parameters?
```
```suggestion
type := "u!"? identifier anchor? urn_anchor? nullability? parameters?
```

As above - I think we can drop quoted name support, as it isn't really supported here.

And anyway, let's try and be consistent within this doc: `name` is defined above as `identifier / quoted_name`, so no need to change that; and also, in EBNF the right symbol is `/`, not `|`. This doc is meant to be a clear EBNF grammar for the user; the syntax should match the Pest grammar, but the way they are described should fit this documentation, not simply repeat the Pest grammar.

GRAMMAR.md
```diff
@@ -213,13 +213,12 @@ The type syntax in this grammar follows the [standard Substrait type definition
...
 ```
 
 Where:
 
-- **`"u!"`** - Optional prefix for user-defined types
-- **`name`** - Type name (case-insensitive, lowercase preferred)
+- **`identifier | quoted_name`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
```
wackywendell: Same, see above

GRAMMAR.md
```diff
@@ -369,13 +368,22 @@ Root[result]
 
 #### Syntax
 
-`function_call := compound_name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
+`function_call := compound_function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
```
```suggestion
`function_call := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type`
```

Let's be consistent here in the naming

wackywendell (changes requested):
Thanks, this is much closer to the behavior I was hoping for. I think the important `u!` behavior is now mostly in place: function base names stay bare, function argument signatures can use `u!` for UDT arguments, and type declarations/references have a normalization story.

I think there are still three main things to clean up:

1. Quoted extension names: we should be consistent here - either allow them consistently (for types and functions, references and extensions), or disallow. Since we're pretty close to disallowing, let's just disallow.
2. Validation: `u!` is only allowed for types; function signatures only allowed for functions.
3. Docs: The `GRAMMAR.md` file should clearly state the rules, in a self-consistent, EBNF form - not simply reiterate the Pest grammar.

See comments below.

GRAMMAR.md
```diff
@@ -213,12 +213,12 @@ The type syntax in this grammar follows the [standard Substrait type definition
...
 
 ```text
-type := ("u!")? (identifier | quoted_name) anchor? urn_anchor? nullability? parameters?
+type := ("u!")? identifier anchor? urn_anchor? nullability? parameters?
 ```
 
 Where:
 
-- **`identifier | quoted_name`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
+- **`identifier`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `"my type"`. Both `u!json` and `json` refer to the same extension; the `u!` prefix is stripped at storage time.
```
wackywendell: 1. `"my type"` is a quoted name, we don't support that anymore
2. Let's note that `u!` is accepted but not recommended

```suggestion
- **`identifier`** - The type name (case-insensitive, lowercase preferred), e.g. `geo_point` or `my_type`. `u!my_type` syntax is accepted, but not recommended; the `u!` will be dropped - e.g. both `u!json` and `json` refer to the same extension.
```

GRAMMAR.md
wackywendell: Can we apply this?

src/parser/extensions.rs
```diff
@@ -240,6 +240,56 @@ impl ParsePair for SimpleExtensionDeclaration {
     }
 }
 
+/// Like [`SimpleExtensionDeclaration::parse_str`] but retains the Pest span of the name
+/// token so that kind-specific validation errors carry a column pointer. Parsing is
+/// duplicated rather than delegated because `parse_pair` discards the span before returning.
+fn parse_decl_for_kind(
```
wackywendell: This makes sense in general shape, but now has a couple new issues:

1. (quibble) We now have both this function and the `SimpleExtensionDeclaration::parse_pair` / `SimpleExtensionDeclaration::from_str` implementations, and they duplicate the same parsing logic. 
2. (blocking) The path used by `substrait-explain` for real extension sections is this one (`parse_decl_for_kind`), while several direct declaration tests still use `SimpleExtensionDeclaration::from_str`. That means those tests are no longer testing the path users actually hit 😬
3. (quibble) This might be nicer as a method, `SimpleExtensionDeclaration::parse_from_kind(s: &str, kind: ExtensionKind) -> Result<Self, MessageParseError>`.

My suggestion:

1. Change this to a method:

   ```rust
   impl SimpleExtensionDeclaration {
       fn parse_from_kind(s: &str, kind: ExtensionKind) -> Result<Self, MessageParseError> {
           ...
       }
   }
   ```

2. Delete the `impl ParsePair` / `FromStr` implementation for `SimpleExtensionDeclaration`; we are not using them in production anymore.
3. Update the direct declaration tests to use `parse_from_kind`, or preferably test through `Parser::parse` where section context matters.

wackywendell (changes requested):
Almost there, a last few things to clean up!

src/parser/extensions.rs
```diff
@@ -700,12 +707,128 @@ Type Variations:
...
+        assert!(
+            Parser::parse(plan_text).is_err(),
+            "function signature suffix on a type declaration should be rejected"
+        );
+    }
+
+    #[test]
+    fn test_u_prefix_function_rejected_at_parser_level() {
+        // The u! rejection must come from parse_decl_for_kind (a MessageParseError with
+        // column info), not from add_extension (an InsertError without position).
```
wackywendell: nit: This comment references a non-existent function now (`parse_decl_for_kind`)

```suggestion
        // The parser should reject function names with a `u!` prefix
```

wackywendell (approved):
LGTM! One small comment change, otherwise good to go!

Issue #162, opened 2026-06-24
[FEATURE] Support multi-line virtual tables for ease of reading


eliot1480:
## Summary

When it comes to representing large virtual tables with substrait-explain, it can sometimes get a little visually cluttered when everything has to be displayed on a single line. If we implemented a format to allow for every row of data to have it's own line, it'd be much easier to read.

## Motivation

By having every row of data have it's own line instead of a single massive line, it makes things much easier to read

## Proposed Solution
When converting from substrait-explain to protobuf, instead of parsing the entire massive table, we parse the rows line by line. And when we convert from protobuf to substrait-explain, we print the rows within the virtual tables line by line.

## Additional Context
Going from this:
```
== Query
=== Plan
Root[id, name, age]
  Sort[($1, &DescNullsFirst) => $0, $1, $2]
    Read:Virtual[(1:i32, 'Alice':string?, 25:i32), (15:i32, 'Bob':string?, 30:i32), (5:i32, null:string?, 20:i32) => id:i32, name:string?, age:i32]
```

To this:
```
== Query
=== Plan
Root[id, name, age]
  Sort[($1, &DescNullsFirst) => $0, $1, $2]
    Read:Virtual[
    (1:i32, 'Alice':string?, 25:i32), 
    (15:i32, 'Bob':string?, 30:i32), 
    (5:i32, null:string?, 20:i32) 
    => id:i32, name:string?, age:i32]
```

Is much more readable.

There are several formats that can be done though:
```
== Query
=== Plan
Root[id, name, age]
  Sort[($1, &DescNullsFirst) => $0, $1, $2]
    Read:Virtual[
      (1:i32, 'Alice':string?, 25:i32), 
      (15:i32, 'Bob':string?, 30:i32), 
      (5:i32, null:string?, 20:i32) 
      => id:i32, name:string?, age:i32]
```

wackywendell:
See also #97 and #105.

The proposed solution makes sense to me: additional lines enclosed by the `[` is pretty reasonable; the alternative in  #105 is also possible, but perhaps not preferred.

Some notes:

[...]

PR #165, opened 2026-06-25
feat!: Add - as a new line type to substrait-explain's parser within virtual tables

eliot1480:
## Description

Added parsing and error handling for multi-line virtual tables. Also added tests to verify the new format.

## Type of Change
New feature. The next steps after this implementation would be to output multi rows for virtual tables that are being converted from substrait, and to update Grammar.md with my changes.

## Testing

- [x] All existing tests pass

## Checklist

- [x] Documentation updated if needed (Need to update grammar.md on completion)
- [x] Pre-commit hooks pass

## Related Issues

Closes #(issue number)
https://github.com/DataDog/substrait-explain/issues/162

### DIFF
--- a/src/parser/expression_grammar.pest
+++ b/src/parser/expression_grammar.pest
@@ -224,9 +224,11 @@ named_column_list = { (named_column ~ ("," ~ sp ~ named_column)*)? }
 planNode = {
     // Specialized read relations must precede read_relation: all start with
     // "Read", and PEG tries alternatives in order — the longer prefixes
-    // must be attempted first.
+    // must be attempted first. virtual_read_relation (full inline form) must
+    // precede virtual_read_relation_open (opening bracket only).
     extension_read_relation
   | virtual_read_relation
+  | virtual_read_relation_open
   | read_relation
   | filter_relation
   | project_relation
@@ -254,10 +256,13 @@ extension_read_relation = { "Read:Extension" ~ "[" ~ named_column_list ~ "]" }
 
 // VirtualTable read: Read:Virtual[rows => columns] or Read:Virtual[_ => columns]
 // Rows are parenthesized tuples; _ means empty (no rows).
-virtual_read_relation = { "Read:Virtual" ~ "[" ~ virtual_read_args ~ sp ~ "=>" ~ sp ~ named_column_list ~ "]" }
-virtual_read_args     = { virtual_row_list | empty }
-virtual_row_list      = { virtual_row ~ (sp ~ "," ~ sp ~ virtual_row)* }
-virtual_row           = { "(" ~ sp ~ expression_list? ~ sp ~ ")" }
+// virtual_read_relation_open is the opening bracket only; rows and schema follow
+// on `- ` continuation lines.
+virtual_read_relation      = { "Read:Virtual" ~ "[" ~ virtual_read_args ~ sp ~ "=>" ~ sp ~ named_column_list ~ "]" }
+virtual_read_relation_open = { "Read:Virtual[" }
+virtual_read_args          = { virtual_row_list | empty }
+virtual_row_list           = { virtual_row ~ (sp ~ "," ~ sp ~ virtual_row)* }
+virtual_row                = { "(" ~ sp ~ expression_list? ~ sp ~ ")" }
 
 filter_relation = { "Filter" ~ "[" ~ expression ~ sp ~ "=>" ~ sp ~ reference_list ~ "]" }
 reference_list  = { (reference ~ (sp ~ "," ~ sp ~ reference)*)? }
--- a/src/parser/structural.rs
+++ b/src/parser/structural.rs
@@ -59,12 +59,29 @@ pub struct Addendum<'a> {
     pub line_no: i64,
 }
 
+/// A `-`-prefixed continuation line attached to a `Read:Virtual` relation.
+/// `content` is the raw text after the leading `"- "` prefix.
+#[derive(Debug, Clone)]
+pub struct Continuation<'a> {
+    pub content: &'a str,
+    pub line_no: i64,
+}
+
+impl<'a> Continuation<'a> {
+    pub fn context(&self) -> ParseContext {
+        ParseContext::new(self.line_no, format!("- {}", self.content))
+    }
+}
+
 /// A relation node in the plan tree, before conversion to a Substrait proto.
 #[derive(Debug, Clone)]
 pub struct RelationNode<'a> {
     pub pair: Pair<'a, Rule>,
     pub line_no: i64,
     pub addenda: Vec<Addendum<'a>>,
+    /// Continuation lines (`- row`) attached to a `Read:Virtual` relation.
+    /// Always empty for all other relation types.
+    pub continuations: Vec<Continuation<'a>>,
     pub children: Vec<RelationNode<'a>>,
 }
 
@@ -77,7 +94,8 @@ impl<'a> RelationNode<'a> {
     }
 }
 
-/// A parsed plan line: either a relation or a `+`-prefixed addendum line.
+/// A parsed plan line: either a relation, a `+`-prefixed addendum, or a
+/// `-`-prefixed continuation line.
 ///
 /// Classification happens at construction time by inspecting the inner grammar
 /// rule, so downstream code can use standard Rust pattern matching rather than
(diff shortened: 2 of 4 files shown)


src/parser/expression_grammar.pest
```diff
@@ -254,10 +256,13 @@ extension_read_relation = { "Read:Extension" ~ "[" ~ named_column_list ~ "]" }
...
 // VirtualTable read: Read:Virtual[rows => columns] or Read:Virtual[_ => columns]
 // Rows are parenthesized tuples; _ means empty (no rows).
-virtual_read_relation = { "Read:Virtual" ~ "[" ~ virtual_read_args ~ sp ~ "=>" ~ sp ~ named_column_list ~ "]" }
-virtual_read_args     = { virtual_row_list | empty }
-virtual_row_list      = { virtual_row ~ (sp ~ "," ~ sp ~ virtual_row)* }
-virtual_row           = { "(" ~ sp ~ expression_list? ~ sp ~ ")" }
+// virtual_read_relation_open is the opening bracket only; rows and schema follow
+// on `- ` continuation lines.
+virtual_read_relation      = { "Read:Virtual" ~ "[" ~ virtual_read_args ~ sp ~ "=>" ~ sp ~ named_column_list ~ "]" }
+virtual_read_relation_open = { "Read:Virtual[" }
```
wackywendell, line 262: I think this rule is the main design fork I’d avoid. It makes multiline `Read:Virtual[` a separate relation form instead of a layout variant of the existing `virtual_read_relation`.

Could we model this as relation-local layout in the grammar instead? For example, something along these lines:

```pest
relation_line_cont = _{ NEWLINE ~ " "* ~ "- " ~ sp }
// Continuation points, used in VirtualRead.
// TODO: use these for other relations as well.
relation_open      = _{ "[" ~ relation_line_cont? }
relation_comma     = _{ sp ~ "," ~ (sp | relation_line_cont) }
relation_arrow     = _{ (sp | relation_line_cont) ~ "=>" ~ sp }
```

Then `virtual_read_relation` can stay the single owner of the syntax:

```pest
virtual_read_relation = {
    "Read:Virtual" ~ relation_open
  ~ virtual_read_args
  ~ relation_arrow
  ~ named_column_list
  ~ "]"
}

virtual_row_list = {
    virtual_row ~ (relation_comma ~ virtual_row)*
}
```

The exact names/rules may differ, but the important property is that multiline input still routes through the same `virtual_read_relation` rule as inline input. Future relations can then opt into the same `relation_open` / `relation_comma` / `relation_arrow` pieces where their bracket syntax supports this layout.

src/parser/structural.rs
```diff
@@ -59,12 +59,29 @@ pub struct Addendum<'a> {
...
+
 /// A relation node in the plan tree, before conversion to a Substrait proto.
 #[derive(Debug, Clone)]
 pub struct RelationNode<'a> {
     pub pair: Pair<'a, Rule>,
     pub line_no: i64,
     pub addenda: Vec<Addendum<'a>>,
+    /// Continuation lines (`- row`) attached to a `Read:Virtual` relation.
+    /// Always empty for all other relation types.
+    pub continuations: Vec<Continuation<'a>>,
```
wackywendell, line 84: This makes continuations a property of relation nodes, but only one relation type can currently interpret them. I’d try to keep the split cleaner: `structural.rs` should know about physical lines, indentation, sections, and relation spans; `relations.rs` should only know about parsed relation grammar.

Concretely, one way of doing this would be to add in a _new_ layer to separate out the distinction between physical lines (ending after the first `newline`) and a `PlanLine` (representing a plan line with its continuations). Here is how I might approach that:

```rust
struct SourceCursor<'a> {
    // The whole text being parsed.
    text: &'a str,
    // Byte offset for the next physical line to inspect.
    next: PhysicalLine,
}

// Offsets to a physical line
#[derive(Copy)]
struct PhysicalLine<'a> {
    // (Pointer to) the entire text of the file/string being read
    text: &'a str,
    // Offsets relative to the entire text
    start: usize,
    end: usize,
    // Line number in the text, for bookkeeping
    line_no: i64,
}

impl<'a> PhysicalLine<'a> {
    fn as_str(self) -> &'a str {
        self.text[start..end]
    }

    fn next(self) -> Option<Self> {
        let next_end = self.text[end..].find('\n')?;
        Self {
           text,
           start: self.end,
           end: next_end,
           line_no: self.line_no + 1,
       }
   }
}

struct PlanLine<'a> {
    // Indentation level for the logical plan line.
    depth: usize,
    // One or more physical lines, still borrowed from the original input.
    text: &'a str,
    line_no: i64,
}

impl<'a> SourceCursor<'a> {
    fn peek_line(&self) -> Option<PhysicalLine<'a>>;
    fn next_plan_line(&mut self) -> Option<PlanLine<'a>>;
}
```

`next_plan_line` can use `peek_line` as a non-consuming helper, track the span end locally as it sees continuation lines, and only advance `offset` once it knows the full logical plan-line span. The grouping rule can stay structural and simple: after reading the first physical line, include following lines while they are exactly one indent deeper and `continuation == true`.

The resulting `PlanLine.text` is still a single slice of the original input, for example:

```text
Read:Virtual[
  - (1, 'alice'),
  - (2, 'bob')
  - => id:i64, name:string]
```

That keeps the structural layer from knowing that the first two continuation lines are rows and the last one is a schema. It only groups physical lines by indentation and the `- ` marker. The grammar decides whether the resulting span is valid relation syntax, and the semantic parser receives the same Pest pair shape as the inline form.

src/parser/structural.rs
```diff
@@ -554,6 +620,46 @@ impl<'a> RelationParser<'a> {
...
+    /// Read:Virtual[
+    ///   - (1, 'alice'),
+    ///   - (2, 'bob')
+    ///   - => id:i64, name:string]
+    /// ```
+    ///
+    /// Row continuations have the form `(expr, ...) [,]`; the schema continuation
+    /// starts with `=>` and ends with `]`.  All continuation strings are slices of
+    /// the original input, so Pest can borrow them without allocation.
+    fn parse_virtual_read_relation_open<'i>(
```
wackywendell, line 636: This path bypasses checks from the normal `VirtualReadRel` parser. For example, inline `Read:Virtual[...]` rejects input children, but this open-form path builds `ctx.children` and then ignores them, so a nested child relation is accepted and disappears from canonical output.

That is the kind of bug I’m worried about with this second parsing path: every invariant now has to be remembered in two places. The continuation-content parser has a similar drift point around row commas: `trim_end_matches(',')` accepts `- (1),,,`, even though that is not grammar-valid inline syntax.

wackywendell (changes requested):

wackywendell:
Thanks for pushing this forward! I agree with the goal: large `Read:Virtual` rows need a readable multiline form, and the `- ` visual marker works well for that.

I think this should be reworked before merge, though. The main issue here is how we separate parser layers, because that directly affects both correctness and whether this can generalize beyond `Read:Virtual`. Multiline layout is a representation choice in the text format, not new relation semantics. A `Read:Virtual` should mean the same thing, and go through the same semantic parser, whether the author wrote it on one line or several.

The existing `sp` rule is a useful model here: it lets the grammar accept harmless layout variation without making semantic parsers care where spaces appeared. I think multiline relation layout should follow the same pattern. The newline / `- ` marker should be parsed as relation layout in the grammar, at the specific positions where the format wants to allow it, rather than becoming a separate `Read:Virtual` relation path.

In broader terms, this is also about encapsulation, and making the code at each level self-contained, so further code changes (by humans or AIs) can be aware of only the layers they are directly touching:

- parse the text representation into one relation shape at the boundary
- keep syntax/layout decisions in the grammar and structural parser
- keep `relations.rs` focused on relation semantics from Pest pairs

A better shape would be:

- structural parsing uses a cursor over the original input to group a multiline relation span without copying or interpreting row/schema fragments
- the Pest grammar owns the legal multiline layout inside that span
- `relations.rs` keeps parsing relation grammar pairs and stays unaware of physical line continuations

Syntax-wise, I’d suggest allowing continuation lines inside relation brackets only after `[`, after a comma, or before `=>`. That keeps the feature general enough for other relations like `Project` / `Aggregate`, while avoiding “newline marker is whitespace anywhere” behavior.

For example:

```text
Read:Virtual[
  - (1, 'alice'),
  - (2, 'bob')
  - => id:i64, name:string]
```

would parse through the same `virtual_read_relation` path as the inline form, not through a separate `virtual_read_relation_open` path.

eliot1480:
@wackywendell 
I see and totally agree with your feedback on encapsulating the layout layers from the semantics layer. To make sure my changes only remain within the layout layer I propose these changes:

1. Grammar: Add a silent and optional line continuation rule for multi-lines
```
Add a silent layout rule and thread it into virtual_read_relation at exactly the three allowed positions:
// Optional line continuation: a newline, indentation, and the "- " marker.
nl = _{ NEWLINE ~ sp ~ "- " }

virtual_read_relation = {
    "Read:Virtual" ~ "[" ~ nl?              // after "["
    ~ virtual_read_args
    ~ (nl | sp) ~ "=>" ~ sp                 // before "=>"
    ~ named_column_list ~ "]"
}
virtual_row_list = { virtual_row ~ (sp ~ "," ~ (nl | sp) ~ virtual_row)* }  // after each ","

```

2. Structural parser: 
Absorb a relation line that is followed by `- ` continuation lines, those lines are absorbed into span, that is then handed to Pest as a unit rather than line by line

3. Enforce Full span consumption
There is a chance where when `LineNode::parse` calls Pest with `Rule::planNode`, it will only call up to `- ` and not the content after that. So we need to ensure that the entire spani is consumed, marking its termination with (pair.as_span().end = span.len())

4. Maintain Round-trip testing and such

PR #166, opened 2026-06-29
Added chunk logic to support multi-line virtual tables

eliot1480:
## Description

Added support for multi-line virtual tables using a chunking design. We build chunks of lines, and if a line meets a certain criteria, it is either merged or set to become its own independent chunk as we iterate through the substrait-explain plan.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #(issue number)


src/parser/chunks.rs
```diff
@@ -0,0 +1,330 @@
...
+//!
+//! - [`ChunkCursor::peek_line`] — look at the next physical line without
+//!   consuming it,
+//! - [`ChunkCursor::merge`] — extend the current chunk to include that line,
+//! - [`ChunkCursor::next`] — finish the current chunk and produce a cursor for
+//!   the next one.
+//!
+//! ## Why the cursor *is* the chunk
+//!
+//! Rather than a separate "chunk" value plus a cursor that could drift out of
```
wackywendell: > Rather than a separate "chunk" value plus a cursor

AIs have a tendency to talk about plans that are no longer in effect. I would restructure this to remove that, and instead talk just about what it is, rather than what it isn't.

src/parser/chunks.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub fn as_str(&self) -> &'a str {
+        self.line
+    }
+}
+
+/// A forward cursor over the input that yields one chunk at a time.
+///
+/// The cursor represents the chunk currently being built: [`chunk`] is
+/// `&text[..current_end]`, and the unread remainder is `text[current_end..]`.
+/// See the [module docs](self) for the overall model.
```
wackywendell: > `&text[..current_end]`

This is a doc comment on a public struct; but its referring to private attributes. This comment fits on `current_end` (as it already is), but the struct comment should stay focused on the 'public' API of this structure.

src/parser/chunks.rs
```diff
@@ -0,0 +1,330 @@
...
+                // advance `end` past the '\n' to the start of the next line.
+                let content_end = if nl > 0 && rest.as_bytes()[nl - 1] == b'\r' {
+                    nl - 1
+                } else {
+                    nl
+                };
+                Some(Line {
+                    line: &rest[..content_end],
+                    end: self.current_end + nl + 1,
+                })
```
wackywendell:
```suggestion
                let line =(&rest[..content_end]).strip_suffix('\r').unwrap_or(raw_line);
                Some(Line {
                    line,
                    end: self.current_end + nl + 1,
                })
```

This can be simplified with the [`str::strip_suffix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_suffix) std library function

src/parser/expression_grammar.pest
```diff
@@ -256,11 +252,39 @@ read_relation = { "Read" ~ "[" ~ table_name ~ sp ~ "=>" ~ sp ~ named_column_list
...
+// before the newline (after `[`, a row comma, or a row) would leave the parser
+// stranded at the space and reject the relation, unlike the old line-based
+// parser which trimmed each physical line. `relation_line_cont` is tried before
+// the bare `sp` fallback because `sp` matches the empty string and would
+// otherwise shadow it.
+// TODO: reuse these for other bracketed relations (Project, Aggregate, ...).
+relation_line_cont = _{ sp ~ NEWLINE ~ sp ~ "- " ~ sp }
+relation_open      = _{ "[" ~ relation_line_cont? }
+relation_comma     = _{ sp ~ "," ~ (relation_line_cont | sp) }
+relation_arrow     = _{ (relation_line_cont | sp) ~ "=>" ~ sp }
```
wackywendell, line 266: This comment is rather long for the grammar file. Comments here should talk about the grammar, not independent layers like `structural.rs`.

```suggestion
// -- Multi-line relation layout --
// A relation may be written across several lines, with each continuation line
// indented and prefixed by a `- ` marker at defined points in the relation grammar.
//
// TODO: reuse these for other bracketed relations (Project, Aggregate, ...).

// A continuation line transition: newline, followed by spaces and "- "
relation_line_cont = _{ sp ~ NEWLINE ~ sp ~ "- " ~ sp }
// Continuation allowed after relation opening ("["), between arguments (after ","),
// or before the columns (before "=>")
relation_open      = _{ "[" ~ relation_line_cont? }
relation_comma     = _{ sp ~ "," ~ (relation_line_cont | sp) }
relation_arrow     = _{ (relation_line_cont | sp) ~ "=>" ~ sp }
```

src/parser/expression_grammar.pest
```diff
@@ -256,11 +252,39 @@ read_relation = { "Read" ~ "[" ~ table_name ~ sp ~ "=>" ~ sp ~ named_column_list
...
 // VirtualTable read: Read:Virtual[rows => columns] or Read:Virtual[_ => columns]
-// Rows are parenthesized tuples; _ means empty (no rows).
-virtual_read_relation = { "Read:Virtual" ~ "[" ~ virtual_read_args ~ sp ~ "=>" ~ sp ~ named_column_list ~ "]" }
+// Rows are parenthesized tuples; _ means empty (no rows). May also be written
+// across multiple lines using the relation layout rules above:
+//   Read:Virtual[
+//     - (1, 'alice'),
+//     - (2, 'bob')
+//     - => id:i64, name:string]
+virtual_read_relation = { "Read:Virtual" ~ relation_open ~ virtual_read_args ~ relation_arrow ~ named_column_list ~ "]" }
```
wackywendell: Also - I think the standard Pest way of enforcing that a rule [consumes all input](https://pest.rs/book/grammars/syntax.html#start-and-end-of-input) is with an `EOI` (End-of-Input) token:

```suggestion
virtual_read_relation = { "Read:Virtual" ~ relation_open ~ virtual_read_args ~ relation_arrow ~ named_column_list ~ "]" ~ sp ~ EOI }
```

(You may need something else in there like `(sp | newline)* ~ EOI` depending on how newlines are included at the end...)

Then you don't need the `ensure_fully_consumed` function.

src/parser/structural.rs
```diff
@@ -88,6 +89,27 @@ pub enum LineNode<'a> {
...
+///
+/// A relation rule can match a prefix and stop — for example at a stray `- `
+/// continuation line left after the relation's closing `]`. Pest does not
+/// require a rule to reach the end of its input, so without this check that
+/// merged line would be silently dropped from the parsed plan.
+///
+/// Trailing content *on the same physical line* (such as a `// comment`) is
+/// left tolerated, as it always has been. Only unconsumed input that crosses a
+/// newline indicates a dropped continuation line, so that is what we reject.
+fn ensure_fully_consumed(line: &str, line_no: i64, consumed: usize) -> Result<(), ParseError> {
```
wackywendell: See the comment in `expression_grammar.pest` - this is a good thing to enforce, but if we can enforce it more structurally (e.g. in the grammar itself), that will be more robust.

src/parser/structural.rs
```diff
@@ -869,20 +893,59 @@ impl<'a> Parser<'a> {
...
-            if line.trim().is_empty() {
-                self.line_no += 1;
+        let mut cursor = ChunkCursor::new(input, 1);
+        while let Some(c) = cursor {
+            let (chunk, line_no, rest) = self.build_chunk(c);
+            cursor = rest;
+
+            // The chunk slice carries the trailing newline of its last physical
+            // line (see `ChunkCursor`); drop it so single-line chunks match the
+            // old `str::lines()` behaviour exactly.
```
wackywendell: > old `str::lines()` behaviour

This code / comment is attempting to match behavior that no longer exists 😬

Perhaps we should simply document that for consistency, parsed chunks always end with no new lines?
```suggestion
            // For consistency, chunks always end with no newline. Multiline chunks will contain interior newlines.
```

src/parser/structural.rs
```diff
@@ -869,20 +893,59 @@ impl<'a> Parser<'a> {
...
 
+    /// Group the next chunk out of `cursor`: always its first physical line,
+    /// plus any `- ` continuation lines indented one level deeper while we are
+    /// in the plan section.
+    ///
+    /// Grouping the whole relation into one contiguous slice lets the grammar
+    /// see a multi-line relation as a single unit. The [`ChunkCursor`] layer is
+    /// syntax-agnostic, so the `- `/indent policy lives here. Returns the chunk
+    /// text, its first line number, and the cursor for whatever follows.
+    fn build_chunk(&self, mut c: ChunkCursor<'a>) -> (&'a str, i64, Option<ChunkCursor<'a>>) {
```
```suggestion
    /// Group the next chunk out of `cursor`: always its first physical line,
    /// plus any `- ` continuation lines indented one level deeper while we are
    /// in the plan section.
    fn next_chunk(&mut self) -> (&'a str, i64) {
```

Two things:

1. The docstring should focus on what this function did; each function should be explainable in isolation, as its own thing.
2. `next_chunk` is a bit clearer than 'build_chunk` - it suggests fetching the next chunk, rather than building the current.
3. How about we put the `ChunkCursor as an attribute on `self`, and modify it there? Given that its existing state on the parser, it seems like it fits there...

tests/plan_roundtrip.rs
```diff
@@ -749,6 +657,176 @@ Root[result]
     );
 }
 
+// === VirtualTable multi-line (continuation) tests ===
```
wackywendell: 🤔  Can we put these in their own round-trip file? Maybe `multi_line_roundtrip.rs`? `plan_roundtrip.rs` is getting a bit large and unwieldy, might be easier to follow if we split it up.

tests/plan_roundtrip.rs
```diff
@@ -749,6 +657,176 @@ Root[result]
...
+  Read:Virtual[(1, 'alice') => id:i64, name:string]"#;
+
+    let multiline = r#"
+=== Plan
+Root[id, name]
+  Read:Virtual[
+    - (1, 'alice')
+    - => id:i64, name:string]"#;
+
+    assert_roundtrip_canonical(inline.trim(), multiline.trim());
```
wackywendell: 👍

src/parser/structural.rs
```diff
@@ -869,20 +893,59 @@ impl<'a> Parser<'a> {
...
+            // old `str::lines()` behaviour exactly.
+            let chunk = chunk.trim_end_matches(['\r', '\n']);
+            if chunk.trim().is_empty() {
                 continue;
             }
 
-            self.parse_line(line)?;
-            self.line_no += 1;
+            self.line_no = line_no;
+            self.parse_line(chunk)?;
```
wackywendell, line 895:
```suggestion
            // TODO: error messages will confusingly point to both the first line of the chunk, and then also the relative line that Pest sees for the chunk; not sure what the format is. We should address this...
            self.parse_line(chunk)?;
```

Let's either add a TODO here, or take a look at addressing this. Can you get an error message as an example to put here?

E.g. if line 8 in the text has two continuation lines, and the second has an issue, then `line_no == 8`, but then the Pest span I think will have its own `line_no == 3`, and... that could be confusing; the issue is actually at `line_no == 11`...

tests/plan_roundtrip.rs
```diff
@@ -749,6 +657,176 @@ Root[result]
...
+    );
+
+    assert_roundtrip_canonical(inline, multiline);
+}
+
+#[test]
+fn test_virtual_read_multiline_rejects_child() {
+    // A child relation (not a `- ` line) under a multi-line virtual read is not
+    // merged into the chunk; it becomes a child, which a leaf read rejects.
+    // This is the invariant a separate parsing path would have to re-implement.
```
wackywendell:
```suggestion
    // A child relation (not a `- ` line) under a multi-line virtual read is not
    // merged into the chunk; it becomes a child, and Read nodes have no children.
```

```
This is the invariant a separate parsing path would have to re-implement.
```

This comment 😆

tests/plan_roundtrip.rs
```diff
@@ -749,6 +657,176 @@ Root[result]
...
+    assert!(
+        Parser::parse(plan.trim()).is_err(),
+        "child relation under multi-line Read:Virtual should be rejected"
+    );
+}
+
+#[test]
+fn test_virtual_read_multiline_rejects_stray_continuation() {
+    // A `- ` continuation line after the relation closes is merged into the
+    // chunk but left unconsumed by the grammar; it must error, not be dropped.
```
wackywendell: Good test 👍

wackywendell (changes requested):
In general, this structure looks good! I have a bunch of smaller comments, which I hope should be fairly straightforward to address. Don't let the ❌ confuse you, this is a generally positive review with some changes requested!

Along with my in-line comments, can you also update `GRAMMAR.md` to include multi-line syntax?

tests/plan_roundtrip.rs
```diff
@@ -7,39 +7,8 @@
...
-        canonical.trim(),
-        "Canonical did not roundtrip to itself"
-    );
-    assert_eq!(
-        text2.trim(),
-        canonical.trim(),
-        "Equivalent did not roundtrip to canonical"
-    );
-}
+use common::{assert_roundtrip_canonical, roundtrip_plan};
```
wackywendell, line 10: Yes, good move 👍

wackywendell (commented):
LGTM! Great work!

wackywendell (approved):

PR #167, opened 2026-06-30
feat!: Allow multi-row virtual tables to be emitted

eliot1480:
## Description

This PR enables multi-row virtual tables to be emitted when converting from protobuf to substrait-explain. It works by adding an ArgsLayout enum and a layout field to arguments, so we attach how we format arguments to the arguments parameter when it's passed into relations. Whether we want it to be multi-line or in-line can be decided when passing into a relation function.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #(issue number)

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -655,8 +655,10 @@ Root[id, name]
 ```
 
 The multi-line form is purely a layout convenience: it parses to exactly the
-same plan as the inline form above. (Formatting currently always emits the
-inline form.)
+same plan as the inline form above. When formatting a plan back to text, a
+`Read:Virtual` is emitted in this multi-line form once it has at least
+`OutputOptions::virtual_table_multiline_threshold` rows (3 by default), and
+inline below that. Raising the threshold keeps larger tables inline.
 
 ### `ExtensionTable` Read Relation
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -655,10 +655,7 @@ Root[id, name]
 ```
 
 The multi-line form is purely a layout convenience: it parses to exactly the
-same plan as the inline form above. When formatting a plan back to text, a
-`Read:Virtual` is emitted in this multi-line form once it has at least
-`OutputOptions::virtual_table_multiline_threshold` rows (3 by default), and
-inline below that. Raising the threshold keeps larger tables inline.
+same plan as the inline form above. The standard output may use inline or multi-line form depending on the length of the `VirtualTable`.
 
 ### `ExtensionTable` Read Relation
 
--- a/src/textify/foundation.rs
+++ b/src/textify/foundation.rs
@@ -51,6 +51,10 @@ pub struct OutputOptions {
     /// Show the binary values for literal types as hex strings. Normally, they
     /// are shown as '{{binary}}'
     pub show_literal_binaries: bool,
+    /// A `Read:Virtual` with at least this many rows is emitted across multiple
+    /// lines (one `- row` per line) instead of inline. Set to a very large
+    /// value to keep virtual tables inline regardless of row count.
+    pub virtual_table_multiline_threshold: usize,
 }
 
 impl Default for OutputOptions {
@@ -65,6 +69,7 @@ impl Default for OutputOptions {
             nullability: false,
             indent: "  ".to_string(),
             show_literal_binaries: false,
+            virtual_table_multiline_threshold: 3,
         }
     }
 }
@@ -84,6 +89,7 @@ impl OutputOptions {
             nullability: true,
             indent: "  ".to_string(),
             show_literal_binaries: true,
+            virtual_table_multiline_threshold: 3,
         }
     }
 }
--- a/src/textify/foundation.rs
+++ b/src/textify/foundation.rs
@@ -53,7 +53,7 @@ pub struct OutputOptions {
     pub show_literal_binaries: bool,
     /// A `Read:Virtual` with at least this many rows is emitted across multiple
     /// lines (one `- row` per line) instead of inline. Set to a very large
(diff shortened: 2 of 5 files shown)


src/textify/rels.rs
```diff
@@ -358,18 +416,24 @@ impl<'a> Relation<'a> {
...
+                let positional: Vec<Value> = vt
                     .expressions
                     .iter()
                     .map(|row| Value::Tuple(row.fields.iter().map(Value::Expression).collect()))
                     .collect();
 
+                // Emit many rows across multiple lines for readability
+                // based on configurable threshold (default = 3)
+                let arguments =
+                    if positional.len() >= ctx.options().virtual_table_multiline_threshold {
```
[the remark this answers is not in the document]
wackywendell: suggestion(blocking): This is a good point: Normally a `_` is required; is it not when there are 0 arguments with `virtual_table_multiline_threshold: 0`? And with 0 arguments, what is (1) accepted by the parser in terms of multi-line, (2) what will this formatter code do with a threshold of 0, and (3) what do we _think_ the parser / formatter _should_ accept?

My suggestion: for now, with 0 arguments, only output / accept inline 🤷

Let's also add a test case for this.

src/textify/rels.rs
```diff
@@ -228,12 +228,48 @@ impl<'a> Textify for Emitted<'a> {
...
+    pub fn inline(positional: Vec<Value<'a>>, named: Vec<NamedArg<'a>>) -> Self {
+        Arguments {
+            positional,
+            named,
+            layout: ArgsLayout::Inline,
+        }
+    }
+
+    /// A row-per-line argument list (`- arg` per line) used for `Read:Virtual`
+    /// with many rows. There are no named arguments in this form.
```
wackywendell:
```suggestion
    /// A row-per-line argument list (`- arg` per line) used for `Read:Virtual`
    /// with many rows. Currently not enabled for named arguments.
    /// TODO: enable for named arguments as well.
```

This is something we'll need to do at some point; making a 'no named arguments' assumption is reasonable for now, but its good to mark that this is not a reasonable assumption long-term, that its a short-term way of managing the complexity.

src/textify/rels.rs
```diff
@@ -293,9 +329,20 @@ impl Textify for Relation<'_> {
...
+    ///
+    /// ```text
+    /// Read:Virtual[
+    ///   - (1, 'alice'),
+    ///   - (2, 'bob')
+    ///   - => id:i64, name:string]
+    /// ```
+    ///
+    /// Callers are responsible for any newline that follows (either from an addendum or
+    /// from the next child).
```
wackywendell:
```suggestion
    /// Does not write a trailing newline; callers are responsible for any
    /// newline that follows (either from an addendum or from the next child).
```

'Does not write a trailing newline' is nice and direct, let's leave it...

GRAMMAR.md
```diff
@@ -655,8 +655,10 @@ Root[id, name]
 ```
 
 The multi-line form is purely a layout convenience: it parses to exactly the
-same plan as the inline form above. (Formatting currently always emits the
-inline form.)
+same plan as the inline form above. When formatting a plan back to text, a
+`Read:Virtual` is emitted in this multi-line form once it has at least
+`OutputOptions::virtual_table_multiline_threshold` rows (3 by default), and
+inline below that. Raising the threshold keeps larger tables inline.
```
wackywendell: This `GRAMMAR` file is meant to discuss what the expected `GRAMMAR` is, independent of implementation; discussing specific `OutputOptions` and defaults is out-of-scope here.

```suggestion
same plan as the inline form above. The standard output may use inline or multi-line form depending on the length of the `VirtualTable`.
```

src/textify/foundation.rs
```diff
@@ -51,6 +51,10 @@ pub struct OutputOptions {
     /// Show the binary values for literal types as hex strings. Normally, they
     /// are shown as '{{binary}}'
     pub show_literal_binaries: bool,
+    /// A `Read:Virtual` with at least this many rows is emitted across multiple
+    /// lines (one `- row` per line) instead of inline. Set to a very large
+    /// value to keep virtual tables inline regardless of row count.
```
wackywendell:
```suggestion
    /// A `Read:Virtual` with at least this many rows is emitted across multiple
    /// lines (one `- row` per line) instead of inline. Set to a very large
    /// value to keep virtual tables inline regardless of row count. Defaults to 3.
```

Hm, we don't have this written for other options, but I think this is probably the right place to document defaults, along with the option descriptions...

tests/multi_line_roundtrip.rs
```diff
@@ -171,3 +174,50 @@ Root[id]
...
+    - => id:i64, name:string]"#;
+
+    assert_roundtrip_canonical(inline.trim(), multiline.trim());
+}
+
+#[test]
+fn test_virtual_read_threshold_is_configurable() {
+    // Raising the threshold keeps an otherwise multi-line table (3 rows) inline.
+    let plan = Parser::parse(
+        r#"
```
wackywendell: quibble: This is a bit hard to read; the other examples that use `let blah = r#"\n…"#;` and then have separate processing / assertions are easier to read...

wackywendell (commented):
Quite good, thank you for the nice code structure and good comments! I had a couple small suggestions, and Codex has a good point about a blocking one: if `multiline_threshold: 0` and there are 0 arguments, this may not round-trip successfully; we should probably fix that and add a test.

The other comments are pretty small, feel free to use your judgment on them.

wackywendell (approved):

PR #168, opened 2026-07-01
feat: added versioning to plans, and textifies versions from substrait plans

eliot1480:
## Description

Added a version section to substrait-explain, and added support for round-tripping the version from explain to protobuf, and back. The Version field is not required, and versions at 0.0.0 will not be represented unless there is info on the producer/git hash.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x ] Pre-commit hooks pass

## Related Issues

Closes #(issue number)

### DIFF
--- a/src/parser/structural.rs
+++ b/src/parser/structural.rs
@@ -10,7 +10,7 @@ use pest::iterators::Pair;
 use substrait::proto::extensions::AdvancedExtension;
 use substrait::proto::{
     AggregateRel, FetchRel, FilterRel, JoinRel, Plan, PlanRel, ProjectRel, ReadRel, Rel, RelRoot,
-    SortRel, plan_rel,
+    SortRel, Version, plan_rel,
 };
 
 use crate::extensions::any::Any;
@@ -26,6 +26,7 @@ use crate::parser::relations::{ExtensionReadRel, RelationParsingContext, Virtual
 use crate::parser::{ErrorKind, ExpressionParser, RelationParsePair, Rule, unwrap_single_pair};
 
 pub const PLAN_HEADER: &str = "=== Plan";
+pub const VERSION_HEADER: &str = "=== Version";
 
 /// Represents an input line, trimmed of leading two-space indents and final
 /// whitespace. Contains the number of indents and the trimmed line.
@@ -165,6 +166,9 @@ impl<'a> LineNode<'a> {
 pub enum State {
     // The initial state, before we have parsed any lines.
     Initial,
+    // The version section, after parsing the `=== Version` header. Accepts the
+    // optional indented `producer:` / `git_hash:` detail lines.
+    Version,
     // The extensions section, after parsing the header and any other Extension lines.
     Extensions,
     // The plan section, after parsing the header and any other Plan lines.
@@ -812,6 +816,8 @@ pub struct Parser<'a> {
     /// [`next_chunk`](Self::next_chunk). `None` before parsing starts and once
     /// the input is exhausted.
     cursor: Option<ChunkCursor<'a>>,
+    /// The plan version from an optional `=== Version` section
+    version: Option<Version>,
     extension_parser: ExtensionParser,
     extension_registry: ExtensionRegistry,
     relation_parser: RelationParser<'a>,
@@ -861,6 +867,7 @@ impl<'a> Parser<'a> {
             line_no: 1,
             state: State::Initial,
             cursor: None,
+            version: None,
             extension_parser: ExtensionParser::default(),
             extension_registry: ExtensionRegistry::new(),
             relation_parser: RelationParser::default(),
@@ -945,6 +952,7 @@ impl<'a> Parser<'a> {
 
         match self.state {
             State::Initial => self.parse_initial(indented_line),
+            State::Version => self.parse_version(indented_line),
             State::Extensions => self
                 .parse_extensions(indented_line)
                 .map_err(|e| ParseError::Extension(ctx(), e)),
@@ -968,6 +976,13 @@ impl<'a> Parser<'a> {
     fn parse_initial(&mut self, line: IndentedLine) -> Result<(), ParseError> {
         match line {
             IndentedLine(0, l) if l.trim().is_empty() => {}
+            IndentedLine(0, l)
+                if l.strip_prefix(VERSION_HEADER)
+                    .is_some_and(|rest| rest.is_empty() || rest.starts_with(' ')) =>
+            {
+                self.parse_version_header(l)?;
+                self.state = State::Version;
+            }
             IndentedLine(0, simple::EXTENSIONS_HEADER) => {
                 self.state = State::Extensions;
             }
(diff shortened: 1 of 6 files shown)


src/textify/foundation.rs
```diff
@@ -26,6 +26,8 @@ pub enum Visibility {
 /// OutputOptions holds the options for textifying a Substrait type.
 #[derive(Debug, Clone)]
 pub struct OutputOptions {
+    /// Show the `=== Version` section when the plan carries a non-empty version
+    pub show_version: bool,
```
wackywendell: I think we can drop this option, and make it 'always true'; Substrait now mandates a version, and generally substrait-explain should show everything that is semantically in the plan - so it should show the version if its present in the plan.

src/textify/plan.rs
```diff
@@ -54,6 +56,33 @@ impl<'a, E: ErrorAccumulator + Default + Clone> PlanWriter<'a, E> {
...
+        writeln!(
+            w,
+            "{VERSION_HEADER} {}.{}.{}",
+            version.major_number, version.minor_number, version.patch_number
+        )?;
+        if !version.producer.is_empty() {
+            writeln!(w, "{}producer: {}", self.options.indent, version.producer)?;
+        }
+        if !version.git_hash.is_empty() {
+            writeln!(w, "{}git_hash: {}", self.options.indent, version.git_hash)?;
```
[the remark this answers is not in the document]
wackywendell: Hm, that's a good point, and one we should probably handle somewhere... but this already exists elsewhere (e.g. in the `URN`), so that can be a separate issue: https://github.com/DataDog/substrait-explain/issues/169

For purposes here, I think you can ignore or maybe write a comment noting...

src/parser/structural.rs
```diff
@@ -988,6 +1003,104 @@ impl<'a> Parser<'a> {
         Ok(())
     }
 
+    /// Parse the `=== Version <major>.<minor>.<patch>` header line and store
+    /// the resulting [`Version`], leaving `producer` / `git_hash` empty
+    fn parse_version_header(&mut self, line: &str) -> Result<(), ParseError> {
```
```suggestion
    fn parse_version_header(&self, line: &str) -> Result<Version, ParseError> {
```

Generally, immutable methods that return values are preferred where it isn't too complicated; it makes 'the current state' easier to track. Here, it should be fairly straightforward to return a `Version` instead of modifying...

src/textify/plan.rs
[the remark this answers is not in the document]
gord02: If this is applicable, you can add `.trim()` to the strings. This probably isn't super important to do but might as well since it is simple.

tests/plan_roundtrip.rs
```diff
@@ -871,3 +861,123 @@ Root[sum]
 
     roundtrip_plan(plan);
 }
+
+// === Version section tests ===
```
gord02: We can delete these extra comments. They don't add anything. lines 865-868.

src/parser/structural.rs
```diff
@@ -988,6 +1003,104 @@ impl<'a> Parser<'a> {
...
+            minor_number,
+            patch_number,
+            ..Default::default()
+        });
+        Ok(())
+    }
+
+    /// Parses a single line from the version section: either an indented
+    /// `producer:` / `git_hash:` detail line, or a section header
+    fn parse_version(&mut self, line: IndentedLine) -> Result<(), ParseError> {
```
wackywendell: This method is a bit confusing - it says it parses a single line from the version section, but it also returns `Ok` and advances `self.state` if it sees the `EXTENSIONS_HEADER` or `PLAN_HEADER`. Overall, now that we're going from 2 sections to 3, this parser is getting more complicated.

I would suggest instead factoring out a `parse_section_header` method:

1. `parse_line(…)` checks if the line starts with `=== ` and has indent 0. If so, it calls `self.state = self.parse_section_header(self, line)?; return;` before calling any of the others.
2. `parse_section_header` takes `&mut self` and returns `Result<State, ParseError>`
3. Then, we remove the header-checking both from the `parse_version` path and the `parse_extensions` path; its no longer needed. Similarly with `parse_initial` - also not needed; either `parse_section_header` triggers and we continue, or we fail with an unexpected line.
4. Optionally, we encapsulate `Version` in a `VersionParser` like with extensions; it takes lines, mutates its inner state, and when asked returns a `Version` object

Does that make sense? I think that will help clean up this code substantially...

wackywendell (commented):
Generally looks good! A couple changes:

1. An API change - I don't think we should make `show_version` optional - its never valid to leave it off it exists.
2. A bit of refactoring - the current code mixes header parsing with section parsing. It already did that, but going from 2 sections to 3, that gets worse; let's separate that out better.

Overall, though - good work on the parsing and textification!

src/textify/plan.rs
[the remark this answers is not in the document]
eliot1480: I think I'll just go with adding a quick `.trim()` 👍

src/parser/structural.rs
```diff
@@ -971,41 +991,38 @@ impl<'a> Parser<'a> {
...
+        if line == simple::EXTENSIONS_HEADER {
+            return Ok(State::Extensions);
         }
-        Ok(())
+        if line == PLAN_HEADER {
+            return Ok(State::Plan);
+        }
+        if line
+            .strip_prefix(VERSION_HEADER)
+            .is_some_and(|rest| rest.is_empty() || rest.starts_with(' '))
```
```suggestion
```

Do you need this check? I think `self.parse_version_header` would cover it?

I guess the case of something like `=== VersionAndMore` would then get parsed by `parse_version_header`, but I think I'm ok with that; the error message from `parse_version_header` should clarify...

src/parser/structural.rs
```diff
@@ -971,41 +991,38 @@ impl<'a> Parser<'a> {
...
-                        pest::Span::new(l, 0, l.len()).expect("Invalid span?!"),
-                        format!("Unknown initial line: {l:?}"),
-                    ),
-                ));
-            }
+    /// Parse a `===`-prefixed section header line (at indent 0), returning
+    /// the [`State`] to transition into. Also performs any header-specific
+    /// parsing, e.g. storing the plan [`Version`] from a `=== Version`
+    /// header.
+    fn parse_section_header(&mut self, line: &str) -> Result<State, ParseError> {
```
wackywendell, line 1026: Code structure looks good here!

suggestion: I think we should consider the intentionality here about section ordering - are they allowed in any order (as this code now seems to accept), or do we require a specific order (as we did previously)? We should probably document this in the `GRAMMAR.md` file and make sure the code matches the docs.

🤔  I think a standard output order of `Version`, `Extensions`, `Plan` makes sense, but for parsing... hm, a strict order seems unnecessary. I guess I could go either two ways:

1. Allow any order (as here), its fine
2. `Plan` must be last; otherwise, `Version` and `Extensions` can be in any order

What do you think?

wackywendell (approved):
LGTM! A couple small non-blocking comments.

src/parser/structural.rs
eliot1480: I think we should enforce ordering here. My biggest fear would be duplicate headers or such, and it honestly just makes sense to have a concrete order. I can't think of a valid reason behind why someone would want to mix up the order of version -> extensions -> plan anyway.

Issue #23, opened 2025-07-09
Implement Set relations (Union, Intersect, Except)


wackywendell:
## Summary

Implement support for logical Set relations (Union, Intersect, Except) in both parser and textifier.

### 1. Parser Support
- Add `set_relation` rule to grammar
- Implement `SetRel` parsing with operation type and inputs
- Add tests for set operations

### 2. Textifier Support
- Add `impl From<&'a SetRel> for Relation<'a>`
- Handle Union, Intersect, Except operation types
- Add to main `From<&'a Rel>` match statement (currently returns `todo!()`)

## Grammar Design

Proposed syntax:
```
Set[&Union => $0, $1]
  Read[table1 => a:i64, b:string]
  Read[table2 => c:i64, d:string]
```

## Acceptance Criteria

- [ ] Parser supports set relation syntax
- [ ] Textifier handles SetRel conversion
- [ ] All set operation types supported (Union, Intersect, Except)
- [ ] Proper handling of ALL vs DISTINCT variants
- [ ] Column type compatibility validation
- [ ] Roundtrip testing works
- [ ] Add comprehensive tests

## Related

Part of expanding relation support. Future work could include:
- Other join types (HashJoin, MergeJoin, NestedLoopJoin)
- Window operations
- Cross products

PR #172, opened 2026-07-06
feat: added support for set relations

eliot1480:
## Description
Added support for representing set relations in substrait-explain, for both parsing and textifying representation. Round-trip and functionality tests are also included, all passing.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Example update

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues
https://github.com/DataDog/substrait-explain/issues/23
Closes Issue #23


src/textify/rels.rs
```diff
@@ -1034,6 +1035,44 @@ impl<'a> Relation<'a> {
...
         }
     }
+
+    fn from_set<S: Scope>(rel: &'a SetRel, ctx: &S) -> Self {
+        let child_refs: Vec<Option<&'a Rel>> = rel.inputs.iter().map(Some).collect();
+        let (children, _total_columns) = Relation::convert_children(child_refs, ctx);
+
+        // Set relation output has the same width as any one of its inputs
+        // (it's a pass-through, not a concatenation like Join).
+        let width = children
```
gord02: this is duplicated work. convert_children() returns the number of input columns so I think this can be skipped.

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
...
 
+### Set Relation
+
+**Syntax**: `"Set" "[" set_op "=>" reference_list "]"`
+
+**Components**:
+
+- `set_op` - Set operation enum with `&` prefix, using Substrait's `SetOp`
+  variant names directly: `&UnionAll`, `&UnionDistinct`,
+  `&IntersectionPrimary`, `&IntersectionMultiset`,
```
gord02: nit: we don't need to list all the variants here.

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
...
+  `&IntersectionMultisetAll`, `&MinusPrimary`, `&MinusPrimaryAll`,
+  `&MinusMultiset`
+- `reference_list` - Comma-separated list of field references for output columns
+
+A `Set` relation combines two or more inputs (written as indented children,
+like any other multi-input relation), which must all share the same output
+schema. Field references map to that common schema, not a concatenation of
```
gord02: joins don't need to be referenced in a comment about sets.

src/parser/relations.rs
```diff
@@ -1757,6 +1842,120 @@ mod tests {
...
+            vec![example_read_relation().into_rel(None)],
+            &[3],
+            None,
+        );
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_parse_set_relation_requires_matching_input_widths() {
+        // 2 children with a total of 5 columns can't split evenly.
```
gord02: nit: this comment could be phrased better or just deleted.

src/parser/relations.rs
```diff
@@ -1757,6 +1842,120 @@ mod tests {
...
+            ],
+            &[2, 3],
+            None,
+        );
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_parse_set_relation_rejects_widths_with_evenly_dividing_sum() {
+        // Regression test: widths [2, 4] sum to 6, which divides evenly by 2
```
gord02: I don't think this is a worthwhile test and is confusingly worded. Where is division done in the code?

src/parser/structural.rs
```diff
@@ -472,6 +476,7 @@ impl<'a> RelationParser<'a> {
             Rule::sort_relation => self.parse_rel::<SortRel>(extensions, registry, ctx),
             Rule::fetch_relation => self.parse_rel::<FetchRel>(extensions, registry, ctx),
             Rule::join_relation => self.parse_rel::<JoinRel>(extensions, registry, ctx),
+            Rule::set_relation => self.parse_set_relation(registry, ctx),
```
gord02, line 506: Is there a reason extensions isn't passed through like for the other relation types?

src/parser/expression_grammar.pest
```diff
@@ -335,6 +336,12 @@ fetch_relation       =  { "Fetch" ~ "[" ~ fetch_args ~ sp ~ "=>" ~ sp ~ referenc
 join_type     = { "&Inner" | "&LeftSemi" | "&LeftAnti" | "&LeftSingle" | "&LeftMark" | "&Left" | "&RightSemi" | "&RightAnti" | "&RightSingle" | "&RightMark" | "&Right" | "&Outer" }
 join_relation = { "Join" ~ "[" ~ join_type ~ sp ~ "," ~ sp ~ expression ~ sp ~ "=>" ~ sp ~ reference_list ~ "]" }
 
+// SetRel: Set[&UnionAll => $0, $1, $2]
+// Note: Longer prefixes must come before shorter ones (e.g., "&IntersectionMultisetAll"
```
wackywendell, line 340: Good comment 👍

src/parser/structural.rs
```diff
@@ -280,6 +282,8 @@ struct RelationContext<'a> {
     line_no: i64,
     children: Vec<Rel>,
     input_field_count: usize,
+    /// Output field count of each child individually
+    child_field_counts: Vec<usize>,
```
```suggestion
    /// Output field count of each child individually
    child_field_counts: Vec<usize>,
```

Good comment 👍 

Can we drop `input_field_count` and replace it with `child_field_counts`? `input_field_count` should just be the sum of `child_field_counts`, right? If so, why are we keeping both? It makes construction slightly more complicated, and it allows more invalid states to be possible.

Note - you could then make it a method - `fn input_field_counts(&self) -> usize { self.child_field_counts.sum() }` - if its still useful.

Optional alternative: turn `children` into `Vec<(Rel, usize)>`, where the relation and its children are in the same list. That's a bit cleaner - it makes even more invalid states impossible.

src/parser/structural.rs
```diff
@@ -507,6 +513,30 @@ impl<'a> RelationParser<'a> {
...
+            children,
+            input_field_count: _,
+            child_field_counts,
+            addenda,
+        } = ctx;
+        let line = pair.as_str().to_string();
+        let advanced_extension = addenda.into_standard_advanced_extension(registry)?;
+
+        parse_set_relation_pair(pair, children, &child_field_counts, advanced_extension)
+            .map_err(|e| ParseError::Plan(ParseContext::new(line_no, line), e))
```
wackywendell: quibble: There's no need to copy the line into a new string on the main path, you can do that only on the error path.

```suggestion
        let line = pair.as_str();
        let advanced_extension = addenda.into_standard_advanced_extension(registry)?;

        parse_set_relation_pair(pair, children, &child_field_counts, advanced_extension)
            .map_err(|e| ParseError::Plan(ParseContext::new(line_no, line.to_string()), e))
```

src/parser/structural.rs
```diff
@@ -507,6 +513,30 @@ impl<'a> RelationParser<'a> {
         }
     }
 
+    /// Handle `Set` relations separately from [`parse_rel`](Self::parse_rel)
```
wackywendell, line 546: Hm, this function has more lines for the function signature and input deconstruction than the implementation, which suggests to me that the complexity is being obscured more than its being encapsulated.

Perhaps merge this with `parse_set_relation_pair` into a single function?

Alternatively, the more radical refactor that I think would make all this easier would be to replace all instances of `children: Vec<Rel>` and `input_field_count` and `child_field_counts` with something like this:

```
   struct ParsedChild {
       rel: Rel,
       field_count: usize,
   }

   struct RelationContext<'a> {
       pair: Pair<'a, Rule>,
       line_no: i64,
       children: Vec<ParsedChild>,
       addenda: Addenda<'a>,
   }
 ```

 Then helper methods:

 ```rust
impl RelationContext {
   /// Total count of input fields to this relation (summed across children)
   fn input_field_count(&self) -> usize { ... }
   fn child_field_counts(&self) -> impl Iterator<Item = usize> { ... }
   fn children(&self) -> impl Iterator<Item = Rel> { ... }
   /// Create a [`ParseContext`] for this relation, with line and location, for error surfacing
   fn as_parse_context(&self) -> ParseContext { ... }
}
```

Then plumb that through. That's a more significant refactor - it touches existing code in a bunch of places - but it might be more useful; I think the current `input_field_counts` was an imperfect, "make-do" solution, and replacing it with something better would improve this code. It would also mean that `parse_set_relation` could be folded into `parse_rel` without special handling.

That said, I won't insist on it; its a bit of scope creep - simply removing `input_field_count` would be acceptable. But I think it would make this code nicer / easier to follow 😉

src/textify/rels.rs
```diff
@@ -1034,6 +1035,44 @@ impl<'a> Relation<'a> {
             children,
         }
     }
+
+    fn from_set<S: Scope>(rel: &'a SetRel, ctx: &S) -> Self {
+        let child_refs: Vec<Option<&'a Rel>> = rel.inputs.iter().map(Some).collect();
+        let (children, _total_columns) = Relation::convert_children(child_refs, ctx);
+
+        // Set relation output has the same width as any one of its inputs
+        // (it's a pass-through, not a concatenation like Join).
```
wackywendell, line 1044:
```suggestion
        // Set relation output has the same width as any one of its inputs
        // (it's a pass-through, not a concatenation like Join).
        // TODO: we may want to validate that all inputs have the same width (and schema, if possible...), and provide a warning if they do not.
```

I don't think we need to do this now (although you could, and add a warning for it via `ctx`), but it might be good to mark with a TODO.

src/parser/relations.rs
```diff
@@ -1757,6 +1842,120 @@ mod tests {
         assert!(result.is_err());
     }
 
+    #[test]
+    fn test_parse_set_relation() {
```
wackywendell: suggestion: There's a bunch of tests in this file, but I'm not sure they're pulling their "weight" relative to simply some whole-plan tests. I would suggest moving all of these to roundtrip tests (and consider making a separate file for them... plan_roundtrip is getting too big...), and just test the whole plan either works as a roundtrip, or errors; that should cover all these cases, and should be both simpler and shorter to write, and easier to read.

src/textify/rels.rs
```diff
@@ -1589,6 +1651,98 @@ Filter[gt($0, 10:i32):boolean => $0, $1]
         );
     }
 
+    #[test]
+    fn test_set_relation_union_all() {
```
wackywendell: This test is really long, hard to follow, and closely tied to internal type shapes; I think this is already covered by the roundtrip tests? Perhaps just drop it?

src/textify/rels.rs
```diff
@@ -1589,6 +1651,98 @@ Filter[gt($0, 10:i32):boolean => $0, $1]
...
+        let (result, errors) = ctx.textify(&rel);
+        assert!(errors.is_empty(), "Unexpected errors: {errors:?}");
+        assert!(
+            result.contains("Set[&UnionAll => $0, $1]"),
+            "Unexpected output: {result}"
+        );
+    }
+
+    #[test]
+    fn test_set_relation_unknown_op() {
```
wackywendell, line 1657: This I think is the only test that isn't easily covered by whole plan tests - keep it 👍

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
 # assert_eq!(plan.relations.len(), 1);
 ```
 
+### Set Relation
+
+**Syntax**: `"Set" "[" set_op "=>" reference_list "]"`
+
+**Components**:
+
+- `set_op` - Set operation enum with `&` prefix, using Substrait's `SetOp`
```
```suggestion
- `set_op` - Set operation enum with `&` prefix, using Substrait's protobuf `SetOp`
```

quibble: Note that Substrait has a spec, and a protobuf implementation; the protobuf is _not_ the spec. So here, we are using the protobuf names; they are not part of the spec. Sorry to be a bit pedantic here 😅

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
...
+- `reference_list` - Comma-separated list of field references for output columns
+
+A `Set` relation combines two or more inputs (written as indented children,
+like any other multi-input relation), which must all share the same output
+schema. Field references map to that common schema, not a concatenation of
+all inputs' fields (unlike `Join`):
+
+- `$0`, `$1`, ... refer to fields of the shared input schema
+
+**Example**:
```
```suggestion
#### Example
```

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
 # assert_eq!(plan.relations.len(), 1);
 ```
 
+### Set Relation
+
+**Syntax**: `"Set" "[" set_op "=>" reference_list "]"`
```
```suggestion
#### Syntax

`"Set" "[" set_op "=>" reference_list "]"`
```

Formatting consistent with the rest of the doc

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
...
+
+**Components**:
```
```suggestion
#### Components
```

Same as above, except in this case, it makes even more sense - this is a header, it should be formatted as a header.

GRAMMAR.md
```diff
@@ -882,6 +882,43 @@ Root[user_orders]
...
+A `Set` relation combines two or more inputs (written as indented children,
+like any other multi-input relation), which must all share the same output
+schema. Field references map to that common schema, not a concatenation of
+all inputs' fields (unlike `Join`):
+
+- `$0`, `$1`, ... refer to fields of the shared input schema
+
+**Example**:
+
+```rust
```
wackywendell, line 905: Good example - makes sense, looks like a coherent plan, and its tested 👍

wackywendell (commented):
I have a bunch of comments, but mostly they are relatively small or non-blocking suggestions for cleanup. I would suggest unifying the `input_field_count` and `child_field_counts`, to not duplicate; I have a suggestion for a more significant refactor, which I leave up to your discretion - it would, I think, leave the code in better shape, but also be a larger change.

src/parser/relations.rs
eliot1480: Division isn't necessarily done within the test itself, it's testing for when you have two incompatible children that can be evenly divided by 2 children, the parser doesn't accept it and throws an error. I will rename the test tho, too wordy 😆

src/parser/structural.rs
eliot1480: Set never uses any extensions in it's syntax, and we could pass in extensions for consistency sake, but there's no point in passing in an empty parameter.

GRAMMAR.md
```diff
@@ -882,6 +882,42 @@ Root[user_orders]
...
+#### Components
+
+- `set_op` - Set operation enum with `&` prefix, using Substrait's protobuf `SetOp`
+  variant names directly
+- `reference_list` - Comma-separated list of field references for output columns
+
+A `Set` relation combines two or more inputs (written as indented children,
+like any other multi-input relation), which must all share the same output
+schema. Field references map to that common schema, not a concatenation of
+all inputs' fields:
```
```suggestion
schema. Field references map to that common schema:
```

"not a concatenation of" 😆  this is AI talking about a design we are no longer doing

src/parser/structural.rs
```diff
@@ -491,14 +523,15 @@ impl<'a> RelationParser<'a> {
...
             addenda,
         } = ctx;
         assert_eq!(pair.as_rule(), T::rule());
         let line = pair.as_str();
         let advanced_extension = addenda.into_standard_advanced_extension(registry)?;
+        let input_field_count: usize = children.iter().map(|c| c.field_count).sum();
+        let input_children = into_rels(children);
 
-        match T::parse_pair_with_context(extensions, pair, children, input_field_count) {
+        match T::parse_pair_with_context(extensions, pair, input_children, input_field_count) {
```
wackywendell, line 537:
```suggestion
        // TODO: pass `ParsedChildren` into `parse_pair_with_context`, instead
        // of this overall 'input_field_count'; for some relations (e.g. a LEFT
        // SEMI JOIN), the field count sum is not the correct count.
        match T::parse_pair_with_context(extensions, pair, input_children, input_field_count) {
```

wackywendell (approved):
Looks good! We still have some cleanup in the future around that `input_field_count`, but this puts us on the right track there, and gets Set relations working, excellent! 👍

A couple small suggestions, nothing big!

Issue #34, opened 2025-07-09
Explicit emit output support


wackywendell:
## Summary

Add support for explicit emit output mapping in the text format.

## Problem

The text format currently shows field references like `$0, $1, $2` but doesn't support explicit emit mappings when output columns are reordered or filtered:

- There is no distinction between an output mapping that uses `Emit` but lists all columns and using `Direct`
- Certain relations like `Project` and `Read` may have expressions (Project) or columns (Read) that are not output and hidden; those shown are shown in output order, not the precise way they were put into the plan

See [`RelCommon.emit`](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/proto/substrait/algebra.proto#L17-L33) for the protobuf definition.

## Goal

Enable explicit, precise specification of output column mappings in the text format.

## Success Criteria

- Parser supports explicit emit syntax
- Textifier can optionally output explicit emit mappings, dependent on `OutputOptions`
- Verbose mode can precisely describe the protobuf and round-trip different emit setups, or different Project setups
- Non-verbose mode continues to show precise semantic meaning, but prioritizes readability and simplicity over exact protobuf mapping; e.g. an explicit `Emit` equivalent to `Direct` may end up as `Direct` after round-tripping; in a `Project`, using an expression with a field reference is semantically equivalent to simply using emit to show that field.
- Roundtrip testing covers both modes: verbose roundtrips should preserve protobuf shape, while non-verbose roundtrips should preserve semantic equivalence where normalization is intentional.

## Additional Context: ReadRel and #157

This also connects to #157, which tracks support for `ReadRel.filter` and `ReadRel.best_effort_filter`.

Once `ReadRel` filters are supported, a read may need to filter on fields that are intentionally not emitted downstream. A simpler related use case is a `ReadRel` whose `base_schema` records the real table schema while `RelCommon.emit` selects only some fields for downstream relations.

Today `Read[table => columns]` conflates the read's base schema with the visible output schema. Explicit emit syntax would let the text format preserve both:

```text
Read[my_table, filter=gt($2, 100) => a:i64, b:string, c:i64 |> $0, $1]
```

In that form, `a`, `b`, and `c` are the read's direct output domain / `base_schema`, while `$0, $1` is the final emitted output mapping.

wackywendell:
## Suggested Solution

[...]

PR #173, opened 2026-07-08
feat: add explicit emit output grammar and parsing support

eliot1480:
## Description

Added explicit output grammar and parsing support. When the parser encounters an emit, it will parse that emit and determine if it is a direct emit or explicit emit using `parse_output_emit`. I will be implementing textifying from substrait in another PR.
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues

Closes #(34)
https://github.com/DataDog/substrait-explain/issues/34

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -639,7 +639,7 @@ Root[a, b]
 # assert_eq!(plan.relations.len(), 1);
 ```
 
-Use `+> ... |>` when the read's base schema records the direct output domain but
+Use `+> ... |>` to specify an Emit / output ordering different from the table's base schema:
 only some fields should flow downstream:
 
 ```rust
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -601,9 +601,10 @@ read_output := ("=>" named_column_list) / ("+>" named_column_list ("|>" referenc
 
 - `table_name := name ("." name)*` - table name, optionally qualified with schema/database
 - `named_column := name ":" type` - column name with type annotation
-- `"=>" named_column_list` - legacy implicit output form. It declares the visible output columns without preserving an explicit `RelCommon.emit_kind`.
-- `"+>" named_column_list` - explicit direct output domain. It preserves `RelCommon.emit_kind` as `Direct`.
-- `"+>" named_column_list "|>" reference_list` - explicit direct output domain followed by an explicit `Emit` mapping over that domain.
+- `named_column_list := (named_column ("," named_column)*)?` - the list of columns and their types to be read from the table `table_name`.
+  - `=>` is used to mean implicit column ordering; for `Read`, this translates to `Direct` column ordering.
+  - When used with `+>` and no `|>`, the `named_column`s are in the expected order of the table, and `Direct` emit is used.
+  - When used with `+> … |>`, the `named_column`s are in the expected order of the table, and the emit order is a Remap specified by `reference_list`.
 
 #### Example
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -523,6 +523,7 @@ Where:
 - **`named_arguments`**: Named arguments (optional)
 - **`=>`**: Separator between arguments and output columns (optional, only present when both arguments and columns are specified)
 - **`columns`**: Output column names and types, or field references for pass-through (all relations specify outputs, but format varies)
+- **`|>`**: An alternative/additional separator introducing an explicit `RelCommon.emit` mapping that is always preserved as `Emit`
 
 #### Example
 
@@ -592,12 +593,13 @@ Root[c, d]           // root with output columns c and d
 
 #### Syntax
 
-`"Read" "[" table_name "=>" (named_column ("," named_column)*)? "]"`
+`"Read" "[" table_name "=>" (named_column ("," named_column)*)? explicit_emit_suffix "]"`
 
 #### Components
 
 - `table_name := name ("." name)*` - table name, optionally qualified with schema/database
 - `named_column := name ":" type` - column name with type annotation
+- `explicit_emit_suffix := ("|>" reference_list)?` - optional explicit `RelCommon.emit` mapping
 
 #### Example
 
@@ -616,6 +618,21 @@ Root[result2]
 #
 # let plan = Parser::parse(plan_text).unwrap();
 # assert_eq!(plan.relations.len(), 2);
+```
+Use `|>` when the read's base schema
+records the real table schema but only some fields should flow downstream:
+
+```rust
+# use substrait_explain::Parser;
+#
+# let plan_text = r#"
+=== Plan
+Root[b, a]
+  Read[my_table => a:i64, b:string, c:i64 |> $1, $0]
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
 ```
 
 ### VirtualTable Read Relation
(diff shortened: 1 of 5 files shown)


src/parser/relations.rs
```diff
@@ -233,6 +233,40 @@ fn parse_emit(reference_list: Pair<Rule>, direct_output_count: usize) -> (EmitKi
...
+    match clause.as_rule() {
+        Rule::direct_emit => {
+            let reference_list = unwrap_single_pair(clause);
+            parse_emit(reference_list, direct_output_count)
+        }
+        Rule::explicit_emit => {
+            let reference_list = unwrap_single_pair(clause);
+            let output_mapping = parse_output_mapping(reference_list);
+            let output_count = output_mapping.len();
+            (EmitKind::Emit(Emit { output_mapping }), output_count)
```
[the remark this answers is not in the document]
eliot1480: This is negligible because I will be adding textifying in another PR.

src/parser/expression_grammar.pest
```diff
@@ -225,6 +225,15 @@ table_name = { (name ~ ("." ~ name)*) }
...
 named_column_list = { (named_column ~ ("," ~ sp ~ named_column)*)? }
 
+// -- Read output clauses --
+// `=>` declares the final visible output columns without preserving a RelCommon.emit distinction.
+// `+>` declares the Read direct output domain. Without `|>`, it is an explicit
+// `|>` the trailing reference list is preserved as an explicit Emit over that direct output domain.
+read_output          = { read_implicit_output | read_direct_output }
+read_implicit_output = { "=>" ~ sp ~ named_column_list }
+read_direct_output   = { "+>" ~ sp ~ named_column_list ~ read_emit_suffix }
+read_emit_suffix     = { (sp ~ "|>" ~ sp ~ reference_list)? }
```
wackywendell:
```suggestion
emit_suffix     = { (sp ~ "|>" ~ sp ~ reference_list)? }
```

While we're only implementing this for `Read` now, that is not our general intention - our intention is for this to be standard syntax useable across all relations.

For this PR, it's fine to only wire it into `Read`, but let's build it in a shape (and with naming) such that it is usable across further relations. Ideally, it would be written so that it both works with `Read` today, and the code written here needs no refactoring to be used with other relations - just the other relations need handling.

In this case, `Read` has specific handling for its direct/implicit columns (`a:i64` is specific to `Read`; others require expressions), but the `emit_suffix` should be general.

src/parser/relations.rs
```diff
@@ -233,6 +233,48 @@ fn parse_emit(reference_list: Pair<Rule>, direct_output_count: usize) -> (EmitKi
...
+                ..Default::default()
+            });
+            Ok((columns, common, output_count))
+        }
+        other => unreachable!("Unexpected rule in read_output: {other:?}"),
+    }
+}
+
+fn parse_read_emit_suffix(suffix: Pair<Rule>) -> Option<(EmitKind, usize)> {
+    assert_eq!(suffix.as_rule(), Rule::read_emit_suffix);
```
wackywendell:
```suggestion
fn parse_emit_suffix(suffix: Pair<Rule>) -> Option<(EmitKind, usize)> {
    assert_eq!(suffix.as_rule(), Rule::emit_suffix);
```

As above - this is (will be) more general than `Read`, so let's treat it as such.

GRAMMAR.md
```diff
@@ -592,12 +592,18 @@ Root[c, d]           // root with output columns c and d
...
+read_output := ("=>" named_column_list) / ("+>" named_column_list ("|>" reference_list)?)
+```
 
 #### Components
 
 - `table_name := name ("." name)*` - table name, optionally qualified with schema/database
 - `named_column := name ":" type` - column name with type annotation
+- `"=>" named_column_list` - legacy implicit output form. It declares the visible output columns without preserving an explicit `RelCommon.emit_kind`.
+- `"+>" named_column_list` - explicit direct output domain. It preserves `RelCommon.emit_kind` as `Direct`.
+- `"+>" named_column_list "|>" reference_list` - explicit direct output domain followed by an explicit `Emit` mapping over that domain.
```
```suggestion
- `named_column_list := (named_column ("," named_column)*)?` - the list of columns and their types to be read from the table `table_name`.
  - `=>` is used to mean implicit column ordering; for `Read`, this translates to `Direct` column ordering.
  - When used with `+>` and no `|>`, the `named_column`s are in the expected order of the table, and `Direct` emit is used.
  - When used with `+> … |>`, the `named_column`s are in the expected order of the table, and the emit order is a Remap specified by `reference_list`.
```

The purpose of `components` is to define grammatically (`term := …`) components of the above section, as well as explain their purpose. Ideally, the `GRAMMAR.md` file should have the full EBNF specification for the substrait-explain grammar in it. As `named_column_list` is not defined elsewhere in this document, it should be defined here.

Secondly, this document defines the Grammar in terms of the Substrait spec, which discusses emits / output ordering. This document does not discuss Protobuf specifics like `RelCommon.emit_kind` - those are, in some sense, another implementation of the Substrait spec that is out of scope for this document.

Also - the definition of `reference_list` should be moved to here from the Filter relation section, as this is now the first place in the document that term is defined.

GRAMMAR.md
```diff
@@ -617,6 +623,36 @@ Root[result2]
...
+=== Plan
+Root[a, b]
+  Read[my_table +> a:i64, b:string]
+# "#;
+#
+# let plan = Parser::parse(plan_text).unwrap();
+# assert_eq!(plan.relations.len(), 1);
+```
+
+Use `+> ... |>` when the read's base schema records the direct output domain but
```
```suggestion
Use `+> ... |>` to specify an Emit / output ordering different from the table's base schema:
```

Let's use more base Substrait / relational algebra terminology and discuss in terms of what is intended (schema from a table), and less Protobuf / substrait-rs terminology ("base schema", "direct output domain").

src/textify/rels.rs
```diff
@@ -234,6 +234,9 @@ pub enum ArgsLayout {
     /// `arg, arg, arg` on a single line.
     #[default]
     Inline,
+    /// `arg +> col, col`, used by named `Read` when the direct output domain
+    /// must be shown separately from an explicit emit mapping.
+    DirectDomain,
```
wackywendell: This should be a separate enum from `ArgsLayout`. You can have an `Inline` layout with implicit or explicit emit, and you can have `Rows` layout with either implicit or explicit emit. That's not happening here because you're only adding `+>`/`|>` for regular `Read` and not for `Read:Virtual`, but we should still build orthogonal structures as orthogonal so extending this to other relations is less challenging.

I would suggest a second, orthogonal enum here:

```rust
enum OutputSyntax {
    /// Implicit output ordering - output columns are re-ordered based on output in the form `=> output_columns`
    Implicit,
    /// Implicit output ordering: output columns are output in direct form, with `|> order` appended if a Remap exists
    Explicit,
}
```

src/textify/rels.rs
```diff
@@ -367,13 +382,35 @@ impl Relation<'_> {
                 }
                 write!(w, "{child_indent}- => {cols}]")
             }
+            Some(args) if args.layout == ArgsLayout::DirectDomain => {
+                let args = ctx.display(args);
+                let cols = ctx.separated(self.columns.iter(), ", ");
+                write!(w, "{indent}{name}[{args} +> {cols}")?;
+                self.write_emit_suffix(ctx, w)?;
```
wackywendell: This should all be part of `Emitted`, not here.

src/textify/rels.rs
```diff
@@ -367,13 +382,35 @@ impl Relation<'_> {
...
             Some(args) => {
                 let args = ctx.display(args);
+                let cols = Emitted::new(&self.columns, self.emit);
+                let cols = ctx.display(&cols);
                 write!(w, "{indent}{name}[{args} => {cols}]")
             }
         }
     }
 
+    fn write_emit_suffix<S: Scope, W: fmt::Write>(&self, ctx: &S, w: &mut W) -> fmt::Result {
```
wackywendell: This should also be be part of `Emitted`, not here.

wackywendell (commented):
The Grammar update makes sense, and scoping it to just `Read` is reasonable, good work there!

That said, even if we're just scoping it to `Read` for this PR, it should be architected as something more general that we can adopt for other relations and maintain clarity and orthogonality around the relevant structures, and I have a few suggestions along those lines.

GRAMMAR.md
```diff
@@ -592,12 +592,21 @@ Root[c, d]           // root with output columns c and d
...
+```
 
 #### Components
 
 - `table_name := name ("." name)*` - table name, optionally qualified with schema/database
 - `named_column := name ":" type` - column name with type annotation
+- `named_column_list := (named_column ("," named_column)*)?` - the list of columns and their types to be read from the table `table_name`.
+  - `=>` is used to mean implicit column ordering; for `Read`, this translates to `Direct` column ordering.
+  - When used with `+>` and no `|>`, the `named_column`s are in the expected order of the table, and `Direct` emit is used.
+  - When used with `+> … |>`, the `named_column`s are in the expected order of the table, and the emit order is a Remap specified by `reference_list`.
```
wackywendell, line 610: Note: IIUC, this is a new mention of `reference_list`, which has not yet been defined in the document. Right now its defined under `Filter` below.

Can we move the definition out of the `Filter` section and into the main "General Relation Grammar" [section](https://github.com/DataDog/substrait-explain/blob/7a6b7eee923ea3322e8d445e9dd84ae944ccaedc/GRAMMAR.md?plain=1#L519-L525), where it is first referenced?

src/textify/rels.rs
```diff
@@ -404,11 +478,17 @@ impl<'a> Relation<'a> {
         match &rel.read_type {
             Some(ReadType::NamedTable(table)) => {
                 let table_name = Value::TableName(table.names.iter().map(|n| Name(n)).collect());
+                let output_syntax = if emit.is_some() {
```
wackywendell, line 487:
```suggestion
                // XXX: For `ReadRel`s, we use `=>` if emit is None, `+>` if
                // `emit` is `Some(Direct)`, and `+> … |>` if emit is
                // `Some(Remap(…))`. However, we ignore the
                // `OutputOptions::show_emit` option, and we haven't yet
                // supported other operators; at some point, we should figure
                // out our policy here and clean this up.
                let output_syntax = if emit.is_some() {
```

Let's note this, and we can point to it in the issue as well - this PR is a good first step, but it will leave things in an unusual state.

wackywendell (approved):
Some small comments, otherwise - approved! Thanks for your help on this!

PR #178, opened 2026-07-10
feat: precision literal implementations

### LINKED ISSUE #49: [FEATURE] Full support for timestamp and precisiontimestamp types
## Summary

Currently, there is some [textification support](https://github.com/DataDog/substrait-explain/blob/b7c4f52368b56a11d4b2e905cc9f04da34c6d21f/src/textify/types.rs#L622) for precisiontimestamp, and some parser support for timestamps, but it doesn't look like there's full bidirectional support for all 3-5 types (_Note: names are case-insensitive_):

- `precision_time<P>`
- `precision_timestamp<P>`
- `precision_timestamp_tz<P>`
- ([deprecated](https://github.com/substrait-io/substrait/blob/95f8f350328f63bf44be63f28eb67680983edbb1/proto/substrait/type.proto#L24-L25)) `timestamp`
- (deprecated) `timestamp_tz`

We should probably support all 5 as long as they are in [the official grammar](https://github.com/substrait-io/substrait/blob/c37d459e93875c33c7b728c3deb7c9bfe5fc7694/grammar/SubstraitLexer.g4#L39-L40), with tests to ensure that they can be both parsed and textified.

## Stretch

Ideally, we could also have an `--upgrade` or `--fix` option, which would automatically convert from `timestamp` to `precision_timestamp` and same for `_tz`. That would be neat.

gord02:
## Description

Adds text-format parsing and textification for Substrait's precisiontimestamp, precisiontimestamptz, and precisiontime literals (values carrying an explicit precision of 0/3/6/9/12 — seconds through picoseconds), which previously had no literal support.

### Details

- Parser (src/parser/expressions.rs): Precision 12 (picoseconds) is rejected at parse time — chrono can't represent it — with a clear error; overflow also errors instead of silently truncating.

- Textify (src/textify/expressions.rs): Formatting is best-effort per the project's design: precision 12 is truncated to nanoseconds and rendered with a warning pushed noting the loss. The emitted type suffix reports the truncated the original 12, so not roundtrippable.

- Grammar (GRAMMAR.md): documents the new precisiontime/precisiontimestamp/precisiontimestamptz<N> syntax and that only precisions 0/3/6/9 are supported.
- Tests (src/types_tests.rs, inline unit tests): round-trip coverage for all supported precisions across all three variants, plus best-effort/error-path coverage for precision 12 and unsupported precisions.

No changes to the legacy (deprecated) time/timestamp literal types, which remain fixed at microsecond precision per the Substrait spec.

## Type of Change
- [x] New feature

## Testing

- [x] All existing tests pass

## Related Issues

Closes #49

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -322,9 +322,22 @@ precision_timestamp_type    := "precisiontimestamp" nullability? "<" integer ">"
 precision_timestamp_tz_type := "precisiontimestamptz" nullability? "<" integer ">"
 ```
 
-The precision parameter supports the following unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. We restrict to this set for literals while the *type* syntax accepts the full Substrait range, `0` through `12`. This is contrary to the Substrait specs where any precision in the range of 0 to 12 is fully supported. 
+The precision parameter follows the Substrait unit convention: `0` means
+seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12`
+picoseconds. Two different subsets apply:
 
-*Literals* only support the subset `0`, `3`, `6`, and `9` because chrono (the underlying date/time library) has no sub-nanosecond resolution, so precision-12 literals cannot be parsed. Textifying a precision-12 value from a protobuf plan is still supported: the *value* is truncated to nanosecond resolution and a truncation warning is emitted, but the declared *type* is preserved as `<12>` (truncating the value does not silently rewrite its type). As a result, such textified output does not round-trip, since precision-12 literals cannot be parsed back.
+- *Types* accept the full Substrait range, any precision from `0` through `12`.
+- *Literals* accept only `0`, `3`, `6`, and `9`.
+
+The literal restriction is a limitation of this implementation, not of
+Substrait, which supports every precision from 0 to 12.
+
+Literals stop at `9` because chrono (the underlying date/time library) has no
+sub-nanosecond resolution, so precision-12 literals cannot be parsed.
+Textifying a precision-12 value from a protobuf plan is still supported: the
+*value* is truncated to nanosecond resolution and a truncation diagnostic is
+emitted, but the declared *type* is preserved as `<12>` (truncating the value
+does not silently rewrite its type).
 
 #### Examples
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -241,9 +241,9 @@ A literal can be an integer, float, boolean, string, or null. Literals may inclu
   - A type annotation is required for `null`
 - **`typed_literal`**` := string ":" type`
   - String literals with type annotations for non-primitive types
-  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`
+  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`, `'2023-01-01T12:00:00.123456789':precisiontimestamp<9>`, `'14:30:45.123456':precisiontime<6>`
 
-All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented.
+All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented.
 
 ## Types
 
@@ -309,6 +309,31 @@ Root[result]
 ### Compound Types
 
 Compound types follow the same syntax as standard Substrait parameterized types.
+
+#### Precision Time And Timestamp Types
+
+Precision time and timestamp types put the nullability marker before the precision parameter.
+
+#### Syntax
+
+```text
+precision_time_type         := "precisiontime" nullability? "<" integer ">"
+precision_timestamp_type    := "precisiontimestamp" nullability? "<" integer ">"
+precision_timestamp_tz_type := "precisiontimestamptz" nullability? "<" integer ">"
+```
+
+The precision parameter follows Substrait's unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. Type syntax accepts values from `0` through `12`; precision literal parsing currently supports `0`, `3`, `6`, and `9`.
+
+#### Examples
+
+```text
+precisiontime<6>
+precisiontime?<6>
+precisiontimestamp<9>
+precisiontimestamp?<9>
+precisiontimestamptz<3>
+precisiontimestamptz?<3>
+```
 
 #### Examples
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -243,7 +243,7 @@ A literal can be an integer, float, boolean, string, or null. Literals may inclu
   - String literals with type annotations for non-primitive types
   - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`, `'2023-01-01T12:00:00.123456789':precisiontimestamp<9>`, `'14:30:45.123456':precisiontime<6>`
 
-All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented.
+All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented. The deprecated `timestamp_tz` literal is also not yet implemented; use `precisiontimestamptz<6>` instead.
 
 ## Types
 
@@ -322,7 +322,9 @@ precision_timestamp_type    := "precisiontimestamp" nullability? "<" integer ">"
 precision_timestamp_tz_type := "precisiontimestamptz" nullability? "<" integer ">"
 ```
 
-The precision parameter follows Substrait's unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. Type syntax accepts values from `0` through `12`; precision literal parsing currently supports `0`, `3`, `6`, and `9`.
+The precision parameter supports the following unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. We restrict to this set for literals while the *type* syntax accepts the full Substrait range, `0` through `12`. This is contrary to the Substrait specs where any precision in the range of 0 to 12 is fully supported. 
+
+*Literals* only support the subset `0`, `3`, `6`, and `9` because chrono (the underlying date/time library) has no sub-nanosecond resolution, so precision-12 literals cannot be parsed. Textifying a precision-12 value from a protobuf plan is still supported, but it is truncated to nanoseconds (precision 9) and emits a truncation warning.
 
 #### Examples
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -324,7 +324,7 @@ precision_timestamp_tz_type := "precisiontimestamptz" nullability? "<" integer "
 
 The precision parameter supports the following unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. We restrict to this set for literals while the *type* syntax accepts the full Substrait range, `0` through `12`. This is contrary to the Substrait specs where any precision in the range of 0 to 12 is fully supported. 
 
-*Literals* only support the subset `0`, `3`, `6`, and `9` because chrono (the underlying date/time library) has no sub-nanosecond resolution, so precision-12 literals cannot be parsed. Textifying a precision-12 value from a protobuf plan is still supported, but it is truncated to nanoseconds (precision 9) and emits a truncation warning.
+*Literals* only support the subset `0`, `3`, `6`, and `9` because chrono (the underlying date/time library) has no sub-nanosecond resolution, so precision-12 literals cannot be parsed. Textifying a precision-12 value from a protobuf plan is still supported: the *value* is truncated to nanosecond resolution and a truncation warning is emitted, but the declared *type* is preserved as `<12>` (truncating the value does not silently rewrite its type). As a result, such textified output does not round-trip, since precision-12 literals cannot be parsed back.
 
 #### Examples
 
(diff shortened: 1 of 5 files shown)


GRAMMAR.md
```diff
@@ -310,6 +310,31 @@ Root[result]
...
+
+#### Syntax
+
+```text
+precision_time_type         := "precisiontime" nullability? "<" integer ">"
+precision_timestamp_type    := "precisiontimestamp" nullability? "<" integer ">"
+precision_timestamp_tz_type := "precisiontimestamptz" nullability? "<" integer ">"
+```
+
+The precision parameter follows Substrait's unit convention: `0` means seconds, `3` milliseconds, `6` microseconds, `9` nanoseconds, and `12` picoseconds. Type syntax accepts values from `0` through `12`; precision literal parsing currently supports `0`, `3`, `6`, and `9`.
```
wackywendell: Could we make the supported literal subset more explicit here? Restricting literals to the common precisions `0`, `3`, `6`, and `9` seems reasonable, but it would help to distinguish that from the full Substrait type range of `0..=12`. It would also be useful to mention that precision-12 protobuf literals textify as precision 9 with a truncation warning.

src/textify/expressions.rs
```diff
@@ -70,45 +110,61 @@ fn days_to_date_string(days: i32) -> String {
...
 }
 
-/// Convert microseconds since Unix epoch to timestamp string
-fn microseconds_to_timestamp_string(microseconds: i64) -> String {
+/// Convert a value in precision units since the Unix epoch, to a timestamp string.
+fn precision_timestamp_to_string(value: i64, precision: i32) -> Option<String> {
+    let duration = duration_from_precision_units(value, precision)?;
     let epoch = DateTime::from_timestamp(0, 0).unwrap().naive_utc();
-    let duration = chrono::Duration::microseconds(microseconds);
     let datetime = epoch + duration;
```
wackywendell: Could we use `checked_add_signed` here and return `None` when the result is outside Chrono’s date range? `duration_from_precision_units` can produce a valid `Duration` whose addition to the epoch overflows `NaiveDateTime`; the `+` implementation panics in that case. Textification should report an invalid/unrepresentable protobuf value rather than panic.

src/textify/expressions.rs
```diff
@@ -70,45 +110,61 @@ fn days_to_date_string(days: i32) -> String {
...
+    } else {
+        formatted
+    })
+}
+
+/// Convert a value in precision units since midnight, to a time-of-day string.
+fn precision_time_to_string(value: i64, precision: i32) -> Option<String> {
+    let duration = duration_from_precision_units(value, precision)?;
+    let midnight = NaiveTime::from_hms_opt(0, 0, 0).unwrap();
+    let time = midnight + duration;
```
wackywendell, line 185: `NaiveTime + Duration` wraps modulo 24 hours, so an invalid precision-time value such as 86,460 seconds silently textifies as `00:01:00`. Since this is a time since midnight rather than a duration, could we validate that the value is within one day and emit a failure token/diagnostic when it is not? That keeps malformed protobuf input visible instead of changing its value.

src/parser/expressions.rs
```diff
@@ -307,32 +360,78 @@ fn parse_date_to_days(date_str: &str, span: pest::Span) -> Result<i32, MessagePa
...
+) -> Result<i64, MessageParseError> {
+    let out_of_range = || {
+        MessageParseError::invalid(
+            "precision_timestamp_out_of_range",
+            span,
+            format!("value is out of range for a {literal_kind} literal at precision {precision}"),
+        )
+    };
+    match precision {
+        0 => Ok(duration.num_seconds()),
```
wackywendell: Could we leave a TODO here to reject fractional components that cannot be represented at the declared precision? For example, `.999` at precision 0 currently reaches `num_seconds()` and silently truncates to the epoch. A parser warning is not available today, so this should eventually become an error rather than a lossy conversion.

GRAMMAR.md
```diff
@@ -241,9 +241,9 @@ A literal can be an integer, float, boolean, string, or null. Literals may inclu
   - A type annotation is required for `null`
 - **`typed_literal`**` := string ":" type`
   - String literals with type annotations for non-primitive types
-  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`
+  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`, `'2023-01-01T12:00:00.123456789':precisiontimestamp<9>`, `'14:30:45.123456':precisiontime<6>`
 
-All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented.
+All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented.
```
wackywendell: Could we document that the deprecated `timestamp_tz` literal is still unsupported, and recommend `precisiontimestamptz<6>` instead? I am fine with leaving the deprecated form unimplemented, but this support summary currently makes that boundary easy to miss.

wackywendell (changes requested):
Overall, this is a good addition: updates GRAMMAR, parser, and textifier, with good round-trip tests!

There's a few places where I think we need to handle out-of-range values; I commented specifically, but generally, we should either support them or error, not silently ignore / truncate / wrap around.

Generally, though, good work, keep it going!

src/textify/expressions.rs
```diff
@@ -134,10 +124,11 @@ fn duration_from_precision_units(value: i64, precision: i32) -> Option<chrono::D
 }
 
 /// Convert a value in precision units since the Unix epoch, to a timestamp string.
+/// Returns `None` if `value` is out of chrono's representable date range.
 fn precision_timestamp_to_string(value: i64, precision: i32) -> Option<String> {
     let duration = duration_from_precision_units(value, precision)?;
     let epoch = DateTime::from_timestamp(0, 0).unwrap().naive_utc();
-    let datetime = epoch + duration;
+    let datetime = epoch.checked_add_signed(duration)?;
```
wackywendell: 👍 , nicely done, good use of `?` as well.

src/textify/expressions.rs
```diff
@@ -57,6 +58,45 @@ fn unimplemented_literal<S: Scope, W: fmt::Write>(
...
+        ctx.push_error(
+            PlanError::invalid(
+                variant,
+                Some("value"),
+                "precision 12 (picoseconds) truncated to nanoseconds; sub-nanosecond precision lost",
+            )
+            .into(),
+        );
+    }
+}
```
wackywendell: Could we make this a pure diagnostic constructor, with `write_precision_literal` owning the `precision == 12` check and `ctx.push_error(...)` call?

The warning is itself an object value; accumulating warnings is part of writing. Keeping those responsibilities separate makes the behavior visible from each function’s signature and reduces the exposure of unrelated state.

```suggestion
/// Returns the diagnostic for truncating a picosecond value to nanoseconds.
fn picosecond_truncation_warning(literal_kind: &'static str) -> PlanError {
    PlanError::invalid(
        literal_kind,
        Some("value"),
        "precision 12 (picoseconds) truncated to nanoseconds; sub-nanosecond precision lost",
    )
}
```

src/textify/expressions.rs
```diff
@@ -57,6 +58,45 @@ fn unimplemented_literal<S: Scope, W: fmt::Write>(
...
+
+fn write_precision_literal<S: Scope, W: fmt::Write>(
+    variant: &'static str,
+    value: i64,
+    precision: i32,
+    to_string: impl FnOnce(i64, i32) -> Option<String>,
```
wackywendell: Could we make the formatting result explicit instead of passing a conversion callback returning `Option<String>`?

`None` currently means several different things: unsupported precision, duration overflow, an unrepresentable timestamp, or a time outside the valid day. `write_precision_literal` cannot distinguish them, so it reports every failure as an unsupported precision, even for supported precisions such as `0` and `6`.

Given that there are multiple forms of error here, a struct to track them would probably make sense, and then you could return that in the relevant places:

```rust
enum PrecisionFormatError {
    UnsupportedPrecision,
    OutOfRange,
}

fn duration_from_precision_units(…) -> Result<chrono::Duration, PrecisionFormatError>;

fn format_precision_time(…) -> Result<String, PrecisionFormatError>;

fn format_precision_timestamp(…) -> Result<String, PrecisionFormatError>;

fn write_precision_literal(ctx: &S, w: &mut W, literal_kind: &'static str, precision: i32, formatted: Result<String, PrecisionFormatError>) -> fmt::Result;

impl PrecisionFormatError {
    fn into_plan_error(self, literal_kind: &'static str, precision: i32) -> PlanError { … }
}
```

The call sites would compute the pure result and pass it to the writer:

```rust
write_precision_literal("PrecisionTime", p.precision, format_precision_time(p.value, p.precision), ctx, w)
```

Then the formatters own conversion and validation, while `write_precision_literal` only maps the result to accumulated diagnostics and best-effort output. This has better separation of concerns: it returns `Result` for operations that can legitimately fail, and the signatures keep invariants understandable from local context. In general: passing functions as arguments considerably increases the complexity of a function signature, and is worth considering whether that complexity is necessary.

src/textify/expressions.rs
```diff
@@ -70,45 +110,57 @@ fn days_to_date_string(days: i32) -> String {
...
-    let datetime = epoch + duration;
-
-    // Format with fractional seconds, then clean up trailing zeros
-    let formatted = datetime.format("%Y-%m-%dT%H:%M:%S%.f").to_string();
+/// Convert a value in precision units since midnight, to a time-of-day string.
+/// Returns `None` if `value` falls outside a single day: `NaiveTime + Duration`
+/// wraps modulo 24 hours, which would otherwise silently misrepresent the value.
+fn precision_time_to_string(value: i64, precision: i32) -> Option<String> {
+    let duration = duration_from_precision_units(value, precision)?;
+    if duration < chrono::Duration::zero() || duration >= chrono::Duration::days(1) {
```
wackywendell: 🤔  Does this properly check negative numbers near zero? It might be worth adding a test for `precision_time_to_string(-1, 12)` to see that errors. (And if it doesn't, you may need to check whether `value < 0`, and not the truncated `duration`...)

wackywendell (changes requested):
Thanks for working through the earlier feedback. The latest update addresses the original review well: timestamp overflow no longer panics, invalid time-of-day values no longer wrap, the support boundaries are documented, and the deferred fractional-precision behavior now has a clear TODO. The tests, docs, and code are well aligned.

I’ve left another round of comments around the structure of precision-literal formatting. This goes somewhat beyond the scope of my original review, and I want to acknowledge that I’m asking for more here. Looking at the error paths again, I saw that the current callback returning `Option<String>` loses important distinctions between unsupported precision and out-of-range values. Since this is new code and there's a few different types of errors here, it would be good to clarify some of these boundaries and make sure the right errors/warnings map to the right behavior.

In the comments, I have some specific suggestions for how to do that: have the pure formatting functions return a typed `Result`, and leave `write_precision_literal` responsible for accumulating diagnostics and writing best-effort output. 

Overall, this is moving in a good direction. The feature coverage and handling of malformed protobuf input are noticeably better after the latest changes, and there's just a bit more to do!

src/parser/expressions.rs
```diff
@@ -307,32 +360,81 @@ fn parse_date_to_days(date_str: &str, span: pest::Span) -> Result<i32, MessagePa
...
+    };
+    // TODO: reject fractional components that can't be represented at `precision`.
+    // E.g. `.999` at precision 0 currently reaches `num_seconds()` and is silently
+    // truncated away rather than raising an error.
+    match precision {
+        0 => Ok(duration.num_seconds()),
+        3 => Ok(duration.num_milliseconds()),
+        6 => duration.num_microseconds().ok_or_else(out_of_range),
+        9 => duration.num_nanoseconds().ok_or_else(out_of_range),
+        _ => unreachable!("precision already validated by check_supported_precision"),
```
wackywendell: The `unreachable!` here depends on all callers having already invoked a separate helper. That makes this function’s precondition non-local, and a future
caller could turn malformed user input into a panic; the writer of such a future caller would have to know details of this implementation, which is not generally a good setup.

Could we either:

1. have `check_supported_precision` return a private checked-precision value (i.e. an `enum` for `SupportedPrecision`), and have `duration_to_precision_units` accept that instead of a raw i32, or
2. move the check into this function - so then the `unreachable!` expectation depends only on the implementation of this function

The first encodes the expectation as code structure - aka "Parse, don't validate". The second simply moves the expectation closer to its enforcement.

src/textify/expressions.rs
```diff
@@ -70,45 +121,73 @@ fn days_to_date_string(days: i32) -> String {
...
+    let formatted = datetime.format("%Y-%m-%dT%H:%M:%S%.f").to_string();
+    Ok(if formatted.contains('.') {
+        formatted
             .trim_end_matches('0')
             .trim_end_matches('.')
             .to_string()
-    }
+    } else {
+        formatted
+    })
```
wackywendell: Hm... this format trimming looks a bit suspect.

I might suggest using the appropriate decimal level for each precision? So use `%.9f` for precision 9, `%.6f` for 6, ..., and no `%f` for precision 0? It might make it a bit clearer in the output.

src/textify/expressions.rs
```diff
@@ -173,25 +268,45 @@ fn write_literal_value<S: Scope, W: fmt::Write>(
...
 /// The type suffix for a literal (e.g., `"i32"`, `"fp64"`, `"date"`).
-///
 /// Returns `None` for unimplemented types whose [`write_literal_value`] already
 /// emitted an error token.
-fn literal_type_suffix(lit: &LiteralType) -> Option<&'static str> {
+fn literal_type_suffix(lit: &LiteralType, nullable: bool) -> Option<Cow<'static, str>> {
+    let q = if nullable { "?" } else { "" };
+    // truncate precision-12 (picosecond) values down to nanoseconds, since chrono can't
+    // represent picoseconds. The suffix must report that same truncated precision (9).
+    let displayed_precision = |precision: i32| if precision == 12 { 9 } else { precision };
```
wackywendell: We should not do this. Yes, we're truncating the value here, but we shouldn't change the _type_ of that value, just because we truncated the value.

src/textify/expressions.rs
```diff
@@ -173,25 +268,45 @@ fn write_literal_value<S: Scope, W: fmt::Write>(
 }
 
 /// The type suffix for a literal (e.g., `"i32"`, `"fp64"`, `"date"`).
-///
 /// Returns `None` for unimplemented types whose [`write_literal_value`] already
 /// emitted an error token.
-fn literal_type_suffix(lit: &LiteralType) -> Option<&'static str> {
+fn literal_type_suffix(lit: &LiteralType, nullable: bool) -> Option<Cow<'static, str>> {
```
wackywendell: This structure previously returned a `'static str`, which was then immediately written into a writer `W`. Now, we are constructing `string`s, with extra allocations, just to then write into a writer at the next layer up. In general, this code uses `write!` to write directly to a writer, rather than `format!` to allocate and construct strings before writing them.

I would suggest instead that this function take a `<W: fmt::Write>` parameter, and do the writing here instead. Then you can have the match return `let (name: &'static str, precision: Option<i32>)`, then write in three stages: `write!(w, "{name}")?`; `if nullable {write!(w, "?")?;}`; and `if let Some(p) = precision {write!(w, "<{p}>")?;}}". That nicely separates out the matching from the writing.

wackywendell (approved):
LGTM! Thanks for addressing the comments!

Issue #174, opened 2026-07-08
Implement CrossJoin relations


eliot1480:
## Summary

Implement support for logical Cross relations in both parser and textifier.

### 1. Parser Support
- Add `cross_join_relation` rule to grammar
- Implement `CrossRel` parsing with operation type and inputs
- Add tests for cross join operations

### 2. Textifier Support
- Add `Relation::from_cross` in `src/textify/rels.rs` (follows `from_join`/`from_set`)
- Output columns = concatenation of left + right inputs (no join-type column dropping, since `CrossRel` has none)
- Wire into `Relation::from_rel` match (replaces the current `Cross` fallback to `Unimplemented`)
- Rename `NamedRelation::name()`'s `Cross` arm to `"CrossJoin"`
- Add tests for basic cross output and reference-list pruning

## Grammar Design

Proposed syntax:
```
Root[result]
  Cross[$0, $1, $2, $3]
    Read[table1 => a:i64, b:string]
    Read[table2 => c:i64, d:string]
```

## Acceptance Criteria

- [ ] Parser supports cross join relation syntax
- [ ] Textifier handles CrossRel conversion
- [ ] All cross join operation types supported (Union, Intersect, Except)
- [ ] Proper handling of ALL vs DISTINCT variants
- [ ] Roundtrip testing works
- [ ] Add comprehensive tests

PR #180, opened 2026-07-13
feat: implemented cross relations

eliot1480:
## Description

Implemented support for CrossJoin relations, for both textifying and generating.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

## Related Issues
Closes #(174)
https://github.com/DataDog/substrait-explain/issues/174


src/textify/rels.rs
```diff
@@ -1075,6 +1076,26 @@ impl<'a> Relation<'a> {
             children,
         }
     }
+
+    fn from_cross<S: Scope>(rel: &'a CrossRel, ctx: &S) -> Self {
+        let (children, total_columns) =
+            Relation::convert_children(vec![rel.left.as_deref(), rel.right.as_deref()], ctx);
```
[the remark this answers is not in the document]
benbellick: @eliot1480 is this a real concern worth checking?

src/parser/relations.rs
```diff
@@ -1842,6 +1903,96 @@ mod tests {
         assert!(result.is_err());
     }
 
+    #[test]
+    fn test_parse_cross_relation() {
```
wackywendell: I think you only need one of these three tests, maybe none; these are rather duplicative with the roundtrip tests, and adding too many fine-grained unit tests can make changing the code itself (e.g. refactoring) more complicated.

src/textify/rels.rs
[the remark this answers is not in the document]
eliot1480: @benbellick 100%, I'm gonna push a change to fix right now 🙏

src/textify/rels.rs
```diff
@@ -1686,6 +1707,82 @@ Filter[gt($0, 10:i32):boolean => $0, $1]
...
+                read_type: Some(ReadType::NamedTable(NamedTable {
+                    names: vec![table.into()],
+                    advanced_extension: None,
+                })),
+            }))),
+        }
+    }
+
+    #[test]
+    fn test_cross_relation() {
```
wackywendell, line 1728: I also think these tests can either be dropped to a single test, or dropped entirely in favor of round-trip tests. They are mostly checking emit handling, and (1) the logic of emit handling is not  `CrossRel`-specific, and (2) is mostly covered by round-trip tests.

tests/plan_roundtrip.rs
```diff
@@ -524,6 +524,55 @@ Root[a, b]
...
+  Cross[$0, $1, $2, $3]
+    Read[table1 => a:i64, b:string]
+    Read[table2 => c:i64, d:string]"#;
+
+    roundtrip_plan(plan);
+}
+
+#[test]
+fn test_cross_relation_uneven_widths_roundtrip() {
+    // Unlike Set, a cross product allows inputs of differing widths.
```
wackywendell:
```suggestion
/// Cross product can handle inputs of differing widths
fn test_cross_relation_uneven_widths_roundtrip() {
```

1. Why are we comparing to `Set` specifically?
2. Let's make this a test docstring

GRAMMAR.md
```diff
@@ -954,6 +954,41 @@ Root[id, name]
...
+
+```rust
+# use substrait_explain::Parser;
+#
+# let plan_text = r#"
+=== Plan
+Root[id, name, order_id, amount]
+  Cross[$0, $1, $2, $3]
+    Read[users => id:i64, name:string]        // Fields $0, $1
+    Read[orders => order_id:i64, amount:i32]  // Fields $2, $3
```
```suggestion
    Read[users => id:i64, name:string]
    Read[orders => order_id:i64, amount:i32]
```

We don't have support for comments in substrait-explain, probably shouldn't put them in. This passes 'by accident' - we parse the `Read[…]`, and then anything past that is ignored, which it probably shouldn't be...

wackywendell (commented):
Some small-scale suggestions, mostly around the testing: too many tests can stifle a codebase, so let's make sure we're doing full API-level tests (like the roundtrip ones), and for unit tests, that we're testing the behavior of the specific function (e.g. if emit parsing/textifying is general across multiple relations, we don't need unit tests for `CrossRel`).

GRAMMAR.md
wackywendell: https://github.com/DataDog/substrait-explain/issues/182

eliot1480: Maybe a worthy ticket 🤔 I think I'll take a look at this later.

tests/plan_roundtrip.rs
```diff
@@ -524,6 +524,55 @@ Root[a, b]
...
+  Cross[$0, $1]
+    Read[table1 => a:i64, b:string]"#;
+
+    let err = Parser::parse(one_child).expect_err("CrossRel with one child should fail");
+    let msg = format!("{err}");
+    assert!(
+        msg.contains("exactly 2 input children") || msg.contains("CrossRel"),
+        "expected CrossRel child-count error, got: {msg}"
+    );
+}
```
benbellick, line 574: This doesn't need to be fixed in this PR because this seems to be the precedent of this file, but I'm a bit surprised to see tests for error conditions in a file titled `plan_roundtrip.rs`. Not exactly a roundtrip test.

src/textify/rels.rs
```diff
@@ -521,7 +519,21 @@ impl<'a> Relation<'a> {
                     inputs += child.emitted();
                     children.push(Some(child));
                 }
-                None => children.push(None),
+                None => {
+                    let token = ctx.failure(FormatError::Format(PlanError::invalid(
```
benbellick: I don't know this codebase so well so please let me know if I'm not understanding something, but I'm surprised this change would be necessary since `CrossRel` isn't the first relation to require children. Is this required to get the implementation of `CrossRel` in? Also is this standard procedure to handle invalid plans by "constructing" an error relation like this?

tests/plan_roundtrip.rs
wackywendell: Yes, agreed - `plan_roundtrip` has become 'all end-to-end tests', unfortunately...

src/textify/rels.rs
wackywendell: I agree - this is an inelegant solution to a problem that pre-exists for this PR - that missing children are silently ignored / not printed. We should probably address this, but I think its reasonably out of scope here.

src/textify/rels.rs
```diff
@@ -521,7 +519,21 @@ impl<'a> Relation<'a> {
...
+                        "Rel",
+                        Some("input"),
+                        "Required input relation expected, None found",
+                    )));
+                    children.push(Some(Relation {
+                        name: Cow::Owned(format!("{token}")),
```
wackywendell: This is pretty ugly and incorrect; the previous behavior of returning `None` would seem to make more sense for this function, and the actual text output code is probably where outputting a token belongs, instead of creating a fake `Relation` just to satisfy the input.

eliot1480: True! It was a crude solution that mixed error reporting and placeholder rendering. I'm just going to undo that fix and contain this PR to cross relations 🙏

src/textify/rels.rs
```diff
@@ -1686,6 +1707,82 @@ Filter[gt($0, 10:i32):boolean => $0, $1]
...
+        let (result, errors) = ctx.textify(&rel);
+        assert!(errors.is_empty(), "Expected no errors, got: {errors:?}");
+        assert!(
+            result.contains("Cross[$0, $1, $2, $3, $4, $5]"),
+            "Expected concatenated cross output, got: {result}"
+        );
+        assert!(
+            result.contains("Read[left_tbl") && result.contains("Read[right_tbl"),
+            "Expected both children to be formatted, got: {result}"
+        );
```
benbellick:
```suggestion
        let expected = r#"
Cross[$0, $1, $2, $3, $4, $5]
  Read[left_tbl => category:string?, amount:fp64?, value:i32?]
  Read[right_tbl => category:string?, amount:fp64?, value:i32?]"#
            .trim_start();
        assert_eq!(result, expected);
```

nit: I may have messed up the spacing here but I think this test would be more readable if we just checked the whole output.

src/textify/rels.rs
```diff
@@ -1686,6 +1707,82 @@ Filter[gt($0, 10:i32):boolean => $0, $1]
...
+        };
+        let rel = Rel {
+            rel_type: Some(RelType::Cross(Box::new(cross))),
+        };
+
+        let (result, errors) = ctx.textify(&rel);
+        assert!(errors.is_empty(), "Expected no errors, got: {errors:?}");
+        assert!(
+            result.contains("Cross[$0, $3]"),
+            "Expected pruned cross output, got: {result}"
```
benbellick: same here

kadinrabo (commented):
LGTM I didn't find anything new after the first round of feedback. I'll defer to Wendell for a final stamp as the textify expert

benbellick (approved):
two minor comments but otherwise looks great! Thanks!

PR #181, opened 2026-07-13
fix: fix aggregate expression outputs

### LINKED ISSUE #175: [BUG] All Expression Support in Aggregates
## Description

Currently, substrait-explain parsing restricts the output of Aggregates to be a list of references and or function calls. This does not align with the protobuf and semantics of substrait. 

## Proposed Solution 
`aggregate_output` defined in the expression_grammar.pest file simplifies to:
```
aggregate_output      = { (expression ~ (sp ~ "," ~ sp ~ expression)*)? }
```
In the parsing we should also validate that all expressions in the output are references to those in the expression_grouping list. Return parsing error otherwise.

gord02:
## Description

Fix Aggregate output/grouping-expression handling

Aggregate relation output items were mishandled in two ways:

1. Parser: a function_call output item was always parsed as a new measure, even when it structurally matched an existing grouping expression; should instead map back to that grouping position, not double as a measure.
2. Textifier: grouping-expression output items were rendered as raw $i field references using the expression's dedup/reference index, instead of resolving through grouping_expressions and printing the actual expression — producing syntactically valid but semantically wrong output whenever the index didn't coincidentally match the field number. Fixed in both the current and deprecated grouping code paths.
3.  non-function-call output expressions (literals, etc.) matching a grouping expression by structural equality are now supported, error messages on unmatched output expressions are asserted in tests.

## Type of Change

- [x] Bug fix

## Testing

- [x] All existing tests pass

## Related Issues

Closes #175

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -460,19 +460,9 @@ count:():i64
 
 ### Aggregate Measures
 
-Aggregate measures are used in the output of Aggregate relations. They can be either field references (to pass through existing fields) or aggregate function calls (to compute aggregates).
+Aggregate measures are used in the output of Aggregate relations to compute aggregates. An aggregate measure is written as a plain `function_call` (see [Function Calls section](#function-calls)) in the output position of an Aggregate relation - the syntax is identical, e.g. `sum($2):i64`, `count($1):i64`, `avg($3):fp64`.
 
-#### Syntax
-
-- `aggregate_measure := name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type` - aggregate function call with optional extension anchors and required output type
-- Field references: `$0`, `$1`, ...
-
-#### Examples
-
-- `sum($2):i64`
-- `count($1):i64`
-- `avg($3):fp64`
-- `$0` (field reference to grouping field)
+This syntax only captures a measure's `function_reference`, `arguments`, and `output_type`. The Substrait `Measure` message also carries a `filter` (a per-measure filter expression, separate from the aggregate function) and the `AggregateFunction` itself has `invocation` (e.g. `DISTINCT`), `phase`, and `sorts` (for ordered aggregates) - none of which can currently be expressed in this text format. Parsing always produces `invocation: UNSPECIFIED`, `phase: UNSPECIFIED`, no `sorts`, and no `filter`, regardless of what the original plan contained.
 
 ### IfThen
 
@@ -832,8 +822,8 @@ Root[result]
 - `grouping_sets := grouping_set_list / expression_list` - can be a list of grouping sets (each parenthesized), or a single unparenthesized list for the common, single-set case
 - `grouping_set_list := grouping_set ("," grouping_set)*`
 - `grouping_set := "(" expression_list ")" / "_"` - a grouping set can be (1) a list of expressions, or (2) `_`, the standard we use for empty lists
-- `aggregate_output := (reference | aggregate_measure) ("," (reference | aggregate_measure))*` - comma-separated list of output items
-- `aggregate_measure` - field references or aggregate function calls. See [Aggregate Measures section](#aggregate-measures)
+- `aggregate_output := expression ("," expression)*` - comma-separated list of output items
+- Each output item is either an aggregate function call (which becomes a new measure), or an expression that must match one of the `grouping_sets` expressions by value - since an Aggregate relation's output schema is always exactly `grouping_expressions + measures`. See [Aggregate Measures section](#aggregate-measures)
 
 #### Example
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -462,7 +462,7 @@ count:():i64
 
 Aggregate measures are used in the output of Aggregate relations to compute aggregates. An aggregate measure is written as a plain `function_call` (see [Function Calls section](#function-calls)) in the output position of an Aggregate relation - the syntax is identical, e.g. `sum($2):i64`, `count($1):i64`, `avg($3):fp64`.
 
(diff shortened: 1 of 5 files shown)


src/textify/rels.rs
```diff
@@ -722,7 +722,18 @@ impl<'a> Relation<'a> {
             for group in &rel.groupings {
                 let mut grouping_set: Vec<Value> = vec![];
                 for i in &group.expression_references {
-                    grouping_set.push(Value::Reference(*i as i32));
+                    let value = match rel.grouping_expressions.get(*i as usize) {
```
gord02, line 725: `i` is an index into `rel.grouping_expressions`, not a field reference into the input schema which is resolved to the actual expression rather than printing it as `$i`.

src/textify/rels.rs
```diff
@@ -786,16 +798,10 @@ impl<'a> Relation<'a> {
...
-                expression_index_map.entry(key.clone()).or_insert_with(|| {
-                    let value = Value::Expression(exp);
-                    expression_list.push(value); // new unique expression found
-                    // mapping the byte encoded expression to its index in the group_expression list
-                    let index = i;
-                    i += 1;
-                    index // is expression returned by this closure and inserted into map
-                });
-                grouping_set.push(Value::Reference(expression_index_map[&key]));
+                if seen_expressions.insert(exp.encode_to_vec()) {
```
gord02, line 801: This is the same bug: Value::Reference(index) was replaced with Value::Expression(exp), the index-tracking machinery that value depended on became unused code.

src/parser/expressions.rs
```diff
@@ -502,6 +502,42 @@ impl ScopedParsePair for Cast {
     }
 }
 
+/// Parses an already-unwrapped inner pair of `Rule::expression`.
+pub(crate) fn parse_expression_inner(
```
wackywendell: A couple things: 
1. `_inner` is not a very good name for a function, it looks like its just the 'inside' of another function and doesn't really tell you what its doing. 
2. By exporting an `_inner` function from one module and importing it in another, the internal complexity of the implementations are being exposed. I think you can just directly use `Expression::parse_pair` on the outer (you may need to clone the `Pair`, but a `Pair` is a lightweight reference so that's fine), which makes it clearer where its called what is happening and preserves the interface boundaries.

src/parser/relations.rs
```diff
@@ -700,35 +700,65 @@ impl RelationParsePair for AggregateRel {
...
+fn parse_aggregate_output(
     extensions: &SimpleExtensions,
     output_pair: Pair<'_, Rule>,
     grouping_expressions: &[Expression],
 ) -> Result<(Vec<aggregate_rel::Measure>, Vec<i32>), MessageParseError> {
     assert_eq!(output_pair.as_rule(), Rule::aggregate_output);
+
+    // every output item is either:
+    // - a function call, which becomes a new aggregate measure, or
+    // - an expression which must already be one of `grouping_expressions`.
```
wackywendell:
```suggestion
    // every output item is either:
    // - an expression which must already be one of `grouping_expressions`,
    // - a new function call, which we assume must be an aggregate measure.
    //
    // While it would probably be best to check if a function call was an
    // aggregate function, the substrait-explain text does not distinguish
    // between aggregate measures and grouping expressions, and the
    // `SimpleExtensions` do not have that information either (neither here in
    // the Registry nor in the actual Protobuf `ExtensionFunction` definition),
    // so we can only check whether it's a previous expression.
```

Just clarifying here - our semantics do not distinguish directly between aggregate measures and grouping expressions, nor can we check the expressions, so we must do so from context...

src/parser/relations.rs
```diff
@@ -700,35 +700,65 @@ impl RelationParsePair for AggregateRel {
...
 
-    for aggregate_output_item in output_pair.into_inner() {
-        let inner_item = unwrap_single_pair(aggregate_output_item);
-        match inner_item.as_rule() {
-            Rule::reference => {
-                let field_index = FieldIndex::parse_pair(inner_item);
-                output_mapping.push(field_index.0);
+    for output_item in output_pair.into_inner() {
+        assert_eq!(output_item.as_rule(), Rule::expression);
+        let span = output_item.as_span();
```
wackywendell:
```suggestion
        let expression = Expression::parse_pair(extensions, output_item.clone())?;
```

If you do this, I don't think you need `parse_expression_inner` anymore...

Its still imperfect, as we end up parsing the same sub-tree again if its a `Measure`, via `Measure::parse_pair`, but that was happening anyway.

GRAMMAR.md
```diff
@@ -460,19 +460,9 @@ count:():i64
...
-- `aggregate_measure := name anchor? urn_anchor? "(" (expression ("," expression)*)? ")" ":" type` - aggregate function call with optional extension anchors and required output type
-- Field references: `$0`, `$1`, ...
-
-#### Examples
-
-- `sum($2):i64`
-- `count($1):i64`
-- `avg($3):fp64`
-- `$0` (field reference to grouping field)
+This syntax only captures a measure's `function_reference`, `arguments`, and `output_type`. The Substrait `Measure` message also carries a `filter` (a per-measure filter expression, separate from the aggregate function) and the `AggregateFunction` itself has `invocation` (e.g. `DISTINCT`), `phase`, and `sorts` (for ordered aggregates) - none of which can currently be expressed in this text format. Parsing always produces `invocation: UNSPECIFIED`, `phase: UNSPECIFIED`, no `sorts`, and no `filter`, regardless of what the original plan contained.
```
```suggestion
This syntax only captures a measure's `function_reference`, `arguments`, and `output_type`. The Substrait `Measure` message also carries a `filter` (a per-measure filter expression, separate from the aggregate function) and the `AggregateFunction` itself has `invocation` (e.g. `DISTINCT`), `phase`, and `sorts` (for ordered aggregates), which are currently unsupported. Parsing always produces `invocation: UNSPECIFIED`, `phase: UNSPECIFIED`, no `sorts`, and no `filter`, regardless of what the original plan contained.
```

Simplifying a bit - from the user perspective, these are unsupported features; whether that's because we haven't implemented it or the syntax can't support it or whatever is irrelevant.

wackywendell (commented):
Generally pretty good! I have one small structural suggestion and a couple comment/docs ones, but pretty good!

wackywendell (approved):

PR #183, opened 2026-07-14
feat: interval day to second support

### LINKED ISSUE #177: [FEATURE] Design convention for multi-field literal types (starting with `IntervalDayToSecond`)
**Summary:**
All existing *implemented* literals are single-value (bool/int/float/string) or map onto a single quoted string (Date/Time/Timestamp), parsed via `string_literal` and disambiguated with a `:type` suffix.

`IntervalDayToSecond` is the first literal that doesn't fit that pattern — its proto shape is:
```proto
message IntervalDayToSecond {
  int32 days = 1;
  int32 seconds = 2;
  oneof precision_mode {
    int32 microseconds = 3 [deprecated = true];
    int32 precision = 4;
  }
  int64 subseconds = 5;
}
```

Options for the convention:
Option 1 — Quoted duration string

Text representation:
'1 02:03:04.5':interval_day

Grammar rule: no new rule — reuses the existing string_literal token (src/parser/expression_grammar.pest:16) plus the existing :type suffix (type rule, line 143). Structure (days/seconds/subseconds) is recovered by Rust string-parsing after the grammar hands over the raw string, same pattern as parse_date_to_days (src/parser/expressions.rs:290).

---
Option 2 — Bracketed field list

Text representation:
interval_day_to_second<1, 2, 3, 4>

Grammar rule:
interval_day_to_second_literal = {
  ^"interval_day_to_second" ~ "<" ~ integer ~ "," ~ sp ~ integer ~ "," ~ sp ~ integer ~ "," ~ sp ~ integer ~ ">"
}
We don't need to disambiguate between microseconds and precision since microseconds would only be zero or 6, and precision allows more options(0,3,6,9,12). Since it is deprecated we do not need to roundtrip back to the that field and can just emit the precision value. 

---
Option 3 — Flat positional integers + enum discriminator

Text representation:
interval_day_to_second(1, 7384, &precision, 500000)

Grammar rule:
interval_precision_mode = { "&precision" | "&microseconds" }
interval_day_to_second_literal = {
  ^"interval_day_to_second" ~ "(" ~ integer ~ "," ~ sp ~ integer ~ "," ~ sp ~ interval_precision_mode ~ "," ~ sp ~ integer ~ ")"
}
Keeps days/seconds/subseconds as plain positional integers (no oneof involved there), but resolves the oneof itself with a closed-set enum token — consistent with how join_type/sort_direction already disambiguate fixed variant sets elsewhere in this grammar (src/parser/expression_grammar.pest:322,335), rather than an ambiguous bare integer flag.

Decision needed on which pattern to adopt, since it will set precedent for `IntervalYearToMonth`, `IntervalCompound`, `Decimal`, and other pending multi-field literals.

---
**I think either 1 or 2 makes the most sense.**

gord02:
## Description

Adds Substrait's IntervalDayToSecond type and literal to the text format, on both the parse and textify sides.

- Type: interval_day (bare, unset precision → treated as 6/microseconds) and interval_day<precision> (explicit precision 0–12), matching Type.IntervalDay in the spec.
- Literal: '<terms>':interval_day, e.g. '5d 3s':interval_day, '123456789ns':interval_day. A literal is 1–3 whitespace-separated terms (d, s, and one of ms/us/ns/ps), each term optionally signed and appearing at most once; the subsecond unit determines the stored precision. Literal type ascription is always bare :interval_day — precision is derived from the string, not a type parameter — to avoid two ways of expressing precision disagreeing with each other.
- Enforces the spec's ±3,650,000-day bound (both individually and combined with seconds), i32 bounds on seconds, and per-precision bounds on subseconds.
- Updates GRAMMAR.md with the interval_day_literal grammar and precision semantics.

## Type of Change
- [x] New feature

## Testing
- [x] All existing tests pass

## Related Issues

Closes #177

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -241,9 +241,49 @@ A literal can be an integer, float, boolean, string, or null. Literals may inclu
   - A type annotation is required for `null`
 - **`typed_literal`**` := string ":" type`
   - String literals with type annotations for non-primitive types
-  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`, `'2023-01-01T12:00:00.123456789':precisiontimestamp<9>`, `'14:30:45.123456':precisiontime<6>`
+  - Examples: `'2023-01-01':date`, `'2023-12-25T14:30:45.123':timestamp`, `'2023-01-01T12:00:00.123456789':precisiontimestamp<9>`, `'14:30:45.123456':precisiontime<6>`, `'5d 3s':interval_day<0>`
 
-All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented. The deprecated `timestamp_tz` literal is also not yet implemented; use `precisiontimestamptz<6>` instead.
+All basic literal types (`integer`, `float`, `boolean`, and `string`) are supported, plus `date`, `time`, `timestamp`, `precisiontime`, `precisiontimestamp`, `precisiontimestamptz`, `interval_day`, and typed null literals. Other Substrait literal types (e.g., `interval_year`, `decimal`, `uuid`) are not yet implemented. The deprecated `timestamp_tz` literal is also not yet implemented; use `precisiontimestamptz<6>` instead.
+
+#### `interval_day` Typed Literals
+
+`interval_day` string literals represent Substrait `IntervalDayToSecond` values.
+The string holds up to three duration terms:
+
+```text
+interval_day_literal := (duration_days (" " duration_seconds)? (" " duration_subseconds)?)
+                      / (duration_seconds (" " duration_subseconds)?)
+                      / duration_subseconds
+duration_days        := "-"? digit+ "d"
+duration_seconds     := "-"? digit+ "s"
+duration_subseconds  := "-"? digit+ subsecond_unit
+subsecond_unit       := "ms" / "us" / "ns" / "ps"
+```
+
+Each term is optional, but at least one is required, and terms appear in
+descending order: days, then seconds, then sub-seconds. Terms are separated by
+exactly one space, with no leading or trailing whitespace. Each term carries its
+own optional sign, matching the separate Substrait fields for days, seconds, and
+sub-seconds.
+
+Sub-second precision comes from the type ascription, not the string, so an
+`interval_day` literal always names its precision: `interval_day<precision>`,
+with `precision` an integer from 0 to 12. A sub-second term's unit must agree
+with that precision - `ms` is precision 3, `us` is 6, `ns` is 9, and `ps` is 12 -
+so there is only one place a value's precision can come from.
+
+Examples:
+
+- `'5d':interval_day<0>`
+- `'4d 5s':interval_day<6>`
+- `'123456789ns':interval_day<9>`
+- `'5d 3s 100ms':interval_day<3>`
+- `'-5d 3s':interval_day<0>`
+- `'5d':interval_day?<6>` (nullable)
+
+Only non-zero components are written on output, since precision travels in the
+type suffix: an interval of 5 days at nanosecond precision is `'5d':interval_day<9>`,
+not `'5d 0ns':interval_day<9>`. An all-zero interval is written `'0s'`.
 
 ## Types
 
@@ -283,6 +323,9 @@ From [official Substrait grammar](https://raw.githubusercontent.com/substrait-io
 - `string`, `binary`
 - `timestamp`, `timestamp_tz`, `date`, `time`
 - `interval_year`, `uuid`
+
+`interval_day` is not in this list: it is parameterized by sub-second precision,
+so it is written as a compound type (see below).
 
 #### Nullability
 
@@ -349,6 +392,11 @@ precisiontimestamp?<9>
 precisiontimestamptz<3>
 precisiontimestamptz?<3>
 ```
+
+`interval_day` takes an integer sub-second precision from 0 to 12 (0 = seconds,
+3 = milliseconds, 6 = microseconds, 9 = nanoseconds, 12 = picoseconds), e.g.
+`interval_day<9>`, `interval_day?<0>`. The parameter is required, as it is for
+every other parameterized type.
 
 #### Examples
 
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -267,10 +267,16 @@ own optional sign, matching the separate Substrait fields for days, seconds, and
 sub-seconds.
 
 Sub-second precision comes from the type ascription, not the string, so an
-`interval_day` literal always names its precision: `interval_day<precision>`,
-with `precision` an integer from 0 to 12. A sub-second term's unit must agree
-with that precision - `ms` is precision 3, `us` is 6, `ns` is 9, and `ps` is 12 -
-so there is only one place a value's precision can come from.
+`interval_day` literal always names its precision: `interval_day<precision>`.
+A literal's precision must be one of 0 (seconds), 3 (milliseconds), 6
+(microseconds), 9 (nanoseconds), or 12 (picoseconds) - the precisions that have
+a unit to write a value in. Unlike `precisiontimestamp` and `precisiontime`
+literals, `interval_day` accepts 12: sub-seconds are stored as a plain integer
+count rather than going through `chrono`.
+
+A sub-second term's unit must agree with the ascribed precision - `ms` is
+precision 3, `us` is 6, `ns` is 9, and `ps` is 12 - so there is only one place a
+value's precision can come from.
 
 Examples:
 
@@ -393,10 +399,11 @@ precisiontimestamptz<3>
 precisiontimestamptz?<3>
 ```
 
-`interval_day` takes an integer sub-second precision from 0 to 12 (0 = seconds,
-3 = milliseconds, 6 = microseconds, 9 = nanoseconds, 12 = picoseconds), e.g.
+`interval_day` takes an integer sub-second precision from 0 to 12, e.g.
 `interval_day<9>`, `interval_day?<0>`. The parameter is required, as it is for
-every other parameterized type.
+every other parameterized type. Writing an `interval_day` *literal* additionally
+requires a precision that has a unit to write values in; see
+[`interval_day` Typed Literals](#interval_day-typed-literals).
 
 #### Examples
 
--- a/src/parser/expression_grammar.pest
+++ b/src/parser/expression_grammar.pest
@@ -57,6 +57,33 @@ reference = { "$" ~ integer }
 // Literal
 literal = { (float | integer | boolean | string_literal | null) ~ (":" ~ sp ~ type)? }
 
+// -- interval_day durations --
+//
+// The contents of an `interval_day` literal string, e.g. "5d", "4d 5s",
+// "5d 3s 100ns". A day term, a seconds term, and a sub-second term, in that
+// order; each is optional, but at least one is required, so each unit can
+// appear at most once.
+//
+// This rule is deliberately not referenced from `literal`: the string is
+// unescaped first, then re-parsed with this rule as the start rule. Terms are
+// separated by exactly one space, with no leading or trailing whitespace -
+// literals don't need the looser whitespace the surrounding grammar allows.
+interval_day_duration = {
+    SOI ~ (
+        duration_days ~ (whitespace ~ duration_seconds)? ~ (whitespace ~ duration_subseconds)?
+      | duration_seconds ~ (whitespace ~ duration_subseconds)?
+      | duration_subseconds
+    ) ~ EOI
+}
+
+duration_days       = { integer ~ "d" }
+duration_seconds    = { integer ~ "s" }
+duration_subseconds = { integer ~ subsecond_unit }
+
+// The sub-second units that correspond to a writable `interval_day` precision:
+// ms = 3, us = 6, ns = 9, ps = 12.
+subsecond_unit = { "ms" | "us" | "ns" | "ps" }
+
 // -- Components for types and functions
 anchor     = { "#" ~ sp ~ integer }
 urn_anchor = { "@" ~ sp ~ integer }
@@ -125,10 +152,14 @@ precision_timestamp_tz_type = { ^"precisiontimestamptz" ~ nullability ~ "<" ~ in
 precision_timestamp_type    = { ^"precisiontimestamp"   ~ nullability ~ "<" ~ integer ~ ">" }
 precision_time_type         = { ^"precisiontime"        ~ nullability ~ "<" ~ integer ~ ">" }
 
+// interval_day takes an integer sub-second precision parameter, like the
+// precision timestamp types above.
+// Example: interval_day<6>, interval_day?<9>
+interval_day_type = { ^"interval_day" ~ nullability ~ "<" ~ integer ~ ">" }
+
 // A compound type expression, composed of a name, optional parameters, and optional nullability.
 // Example: LIST?<fp64?>, MAP<i64, string>
-// TODO: interval_day
-compound_type = { list_type | map_type | struct_type | precision_timestamp_tz_type | precision_timestamp_type | precision_time_type }
+compound_type = { list_type | map_type | struct_type | precision_timestamp_tz_type | precision_timestamp_type | precision_time_type | interval_day_type }
 
 // A user-defined type expression. The u! prefix is optional and ignored for lookup purposes;
 // both "json" and "u!json" resolve to the same extension.
--- a/src/parser/expression_grammar.pest
+++ b/src/parser/expression_grammar.pest
@@ -57,33 +57,6 @@ reference = { "$" ~ integer }
 // Literal
 literal = { (float | integer | boolean | string_literal | null) ~ (":" ~ sp ~ type)? }
 
-// -- interval_day durations --
-//
-// The contents of an `interval_day` literal string, e.g. "5d", "4d 5s",
-// "5d 3s 100ns". A day term, a seconds term, and a sub-second term, in that
-// order; each is optional, but at least one is required, so each unit can
-// appear at most once.
-//
-// This rule is deliberately not referenced from `literal`: the string is
-// unescaped first, then re-parsed with this rule as the start rule. Terms are
-// separated by exactly one space, with no leading or trailing whitespace -
-// literals don't need the looser whitespace the surrounding grammar allows.
-interval_day_duration = {
-    SOI ~ (
-        duration_days ~ (whitespace ~ duration_seconds)? ~ (whitespace ~ duration_subseconds)?
-      | duration_seconds ~ (whitespace ~ duration_subseconds)?
-      | duration_subseconds
-    ) ~ EOI
-}
-
-duration_days       = { integer ~ "d" }
-duration_seconds    = { integer ~ "s" }
-duration_subseconds = { integer ~ subsecond_unit }
-
-// The sub-second units that correspond to a writable `interval_day` precision:
-// ms = 3, us = 6, ns = 9, ps = 12.
-subsecond_unit = { "ms" | "us" | "ns" | "ps" }
-
 // -- Components for types and functions
 anchor     = { "#" ~ sp ~ integer }
 urn_anchor = { "@" ~ sp ~ integer }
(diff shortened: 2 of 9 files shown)


src/parser/expressions.rs
```diff
@@ -363,6 +380,189 @@ fn parse_timestamp_to_microseconds(
...
 }
 
+/// Split a single duration term (e.g. "5d", "-3s", "123ns") into its signed numeric
+/// part and unit suffix.
+fn split_duration_term<'a>(
+    term: &'a str,
+    duration_str: &str,
+    span: pest::Span,
+) -> Result<(&'a str, &'a str), MessageParseError> {
+    let digits_start = if term.starts_with('-') { 1 } else { 0 };
```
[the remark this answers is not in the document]
gord02: This isn't work addressing. I don't think adding + in front of number is a case we should account for.

src/textify/expressions.rs
```diff
@@ -111,6 +112,135 @@ fn microseconds_to_timestamp_string(microseconds: i64) -> String {
...
+fn write_interval_day_to_second_literal<S: Scope, W: fmt::Write>(
+    interval: &substrait::proto::expression::literal::IntervalDayToSecond,
+    ctx: &S,
+    w: &mut W,
+) -> fmt::Result {
+    #[allow(deprecated)]
+    let (precision, subseconds) = match interval.precision_mode {
+        Some(PrecisionMode::Precision(p)) => (p, interval.subseconds),
+        Some(PrecisionMode::Microseconds(us)) => (6, us as i64),
+        None => (0, interval.subseconds),
```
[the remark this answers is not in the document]
gord02: This is a good point but is potentially okay to leave as is. Zero makes the most sense as the default.

After looking into it further, I don't think this is a case worth covering. precision is required in the proto and the specs have no mention of being able to not set a precision.

src/parser/expression_grammar.pest
```diff
@@ -97,12 +97,15 @@ simple_type_name = {
   | "date"
   | "time"
   | "interval_year"
+  | "interval_day"
```
wackywendell: Why is this here? `interval_day` is not a 'simple type', its a [compound type](https://substrait.io/types/type_classes/#compound-types).

src/parser/expression_grammar.pest
```diff
@@ -97,12 +97,15 @@ simple_type_name = {
...
   | "uuid"
 }
 
 // A simple native type expression, composed of a name and optional nullability.
 // Examples: i64, bool?, string
-simple_type = { simple_type_name ~ nullability }
+// The trailing "<" lookahead is rejected so that "interval_day<6>" is left for
+// interval_day_type (below) instead of being swallowed as bare "interval_day".
+simple_type = { simple_type_name ~ nullability ~ !"<" }
```
wackywendell: Let's undo this as well; it shouldn't be here in simple_type.

src/parser/expression_grammar.pest
```diff
@@ -125,10 +128,13 @@ precision_timestamp_tz_type = { ^"precisiontimestamptz" ~ nullability ~ "<" ~ in
 precision_timestamp_type    = { ^"precisiontimestamp"   ~ nullability ~ "<" ~ integer ~ ">" }
 precision_time_type         = { ^"precisiontime"        ~ nullability ~ "<" ~ integer ~ ">" }
 
+// The parameterized interval_day<precision> compound type. Bare "interval_day"
+// (no parameters) is handled by simple_type_name above instead.
```
wackywendell: What is this about a 'bare' "interval_day" type? Why are we supporting that?

I... don't think we should do that. For every other type with parameters that we support, we require the parameters to be written out. I think we should do that here, as well.

Looking at the issue - I see I suggested `'30s':interval_day`; but I think leaving off the `<0>` was accidental there, and I don't know that we want to set a precedent for that.

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
     ))
 }
 
+/// Split a single duration term (e.g. "5d", "-3s", "123ns") into its signed numeric
+/// part and unit suffix.
+fn split_duration_term<'a>(
```
wackywendell: I would suggest using Pest here; make a rule for it in Pest, which will be easier to read and give better error messages when it fails to parse, and then parse the pieces of it.

```pest
// One day term, one subseconds term, or both.
interval_day_duration = {
    SOI ~ (
      duration_day ~ (" " ~ duration_seconds)?
                   ~ (" " ~ duration_subseconds)?
      | duration_seconds ~ (" " ~ duration_subseconds)?
      | duration_subseconds
    ) ~ EOI
}

duration_day = { integer ~ "d" }
duration_seconds = { integer ~ "s" }
duration_subseconds = { integer ~ ("ms" | "us" | "ns") }
```

Note also:
- This allows `d`, `s`, and `ms/us/ns`, in that order. Each is optional, at least one required.
- It only allows a single space between, and no leading/trailing, and no tabs. That's a bit strict, but... seems reasonable? Do you really need any of this?

Then... this can go in the main grammar file, but I would suggest not including it as a sub-rule anywhere; I think that will get complicated - I'd just have the Pest parser parse this again, once you got to this level in the code, `ExpressionParser::parse(Rule::interval_day_duration, unescaped_string)`.

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
...
+            ),
+        ));
+    }
+    Ok((value, unit))
+}
+
+/// Mark a duration term category (e.g. "'d'", "subsecond (ns)") as seen, erroring
+/// if it was already seen. Each unit category (`d`, `s`, one subsecond unit) may
+/// appear at most once in a duration string.
+fn mark_seen(
```
wackywendell: See above - with a Pest grammar like the above, the grammar would cover that.

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
...
+    Ok(())
+}
+
+/// Parse a duration string (e.g. "5d", "4d 5s", "123456789ns", "5d 0ns", "-5d -3s") into
+/// an IntervalDayToSecond.
+fn parse_duration_to_interval_day_to_second(
+    duration_str: &str,
+    span: pest::Span,
+) -> Result<substrait::proto::expression::literal::IntervalDayToSecond, MessageParseError> {
+    use substrait::proto::expression::literal::IntervalDayToSecond;
```
wackywendell:
```suggestion
) -> Result<IntervalDayToSecond, MessageParseError> {
```

`use` goes at the top of the file, not inside a function. Then the signature would be more readable to.

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
...
+            format!("Invalid duration '{duration_str}': duplicate {term_name} term"),
+        ));
+    }
+    *seen = true;
+    Ok(())
+}
+
+/// Parse a duration string (e.g. "5d", "4d 5s", "123456789ns", "5d 0ns", "-5d -3s") into
+/// an IntervalDayToSecond.
+fn parse_duration_to_interval_day_to_second(
```
wackywendell: Why not `impl ParsePair for IntervalDayToSecond`?

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
...
+    }
+
+    let total_seconds = days * 86_400 + seconds;
+    let max_total_seconds = MAX_DAYS * 86_400;
+    if !(-max_total_seconds..=max_total_seconds).contains(&total_seconds) {
+        return Err(MessageParseError::invalid(
+            "duration_parse_format",
+            span,
+            format!(
+                "Invalid duration '{duration_str}': days ({days}) and seconds ({seconds}) combine to {total_seconds}s, which exceeds the maximum of {MAX_DAYS}d ({max_total_seconds}s)"
```
wackywendell:
```suggestion
                "Invalid duration '{duration_str}': cannot exceed the maximum of '{MAX_DAYS}d' (10,000 years)"
```

The `total_seconds` is useful for doing the comparison, but is an implementation detail for users, no need for it to leak out.

src/parser/types.rs
```diff
@@ -192,24 +201,43 @@ fn parse_compound_type(
...
         _ => unimplemented!("{:?}", inner.as_rule()),
     }
 }
 
-fn parse_precision_type(pair: Pair<Rule>) -> Result<Type, MessageParseError> {
-    let rule = pair.as_rule();
-    let mut iter = iter_pairs(pair.into_inner());
-    let nullability = iter.parse_next::<Nullability>();
-    let precision_pair = iter.pop(Rule::integer);
+/// Parse and validate a `0..=12` sub-second precision integer.
```
wackywendell: Don't we have a precision type now? Should this simply be parsing into that?

Using a type like that makes it clear when it has/hasn't been validated, so it happens exactly once.

src/textify/expressions.rs
```diff
@@ -111,6 +112,135 @@ fn microseconds_to_timestamp_string(microseconds: i64) -> String {
...
+    };
+
+    let seconds = interval.seconds as i64;
+    let unit = if precision == 0 {
```
wackywendell: You can do this `if precision == 0 && subseconds != 0` check outside the `let unit`, and that will make the rest of this clearer; the nesting here plus the `else` is less readable.

src/textify/expressions.rs
```diff
@@ -111,6 +112,135 @@ fn microseconds_to_timestamp_string(microseconds: i64) -> String {
...
+                    Some("IntervalDayToSecond.subseconds"),
+                    format!(
+                        "IntervalDayToSecond subseconds ({subseconds}) must be 0 when precision is 0"
+                    ),
+                ))
+            );
+        }
+        None
+    } else {
+        match precision {
```
wackywendell: Do we have a type for `Precision`? Can we use that here? It will make this function more readable.

src/textify/expressions.rs
```diff
@@ -111,6 +112,135 @@ fn microseconds_to_timestamp_string(microseconds: i64) -> String {
...
+                            "IntervalDayToSecond precision {precision} cannot be represented as a duration unit; expected one of 0, 3, 6, 9, or 12"
+                        ),
+                    ))
+                );
+            }
+        }
+    };
+
+    const MAX_DAYS: i32 = 3_650_000;
+    if !(-MAX_DAYS..=MAX_DAYS).contains(&interval.days) {
```
wackywendell: While this is technically invalid substrait, I think you shouldn't refuse to write it - just write it, but _also_ put an error into the error channel (via `ctx.push_error`).

src/textify/expressions.rs
```diff
@@ -111,6 +112,135 @@ fn microseconds_to_timestamp_string(microseconds: i64) -> String {
...
+
+    // `days` alone can be in range while combined with `seconds` it exceeds the
+    // spec's day bound (see the matching check in the parser).
+    let total_seconds = interval.days as i64 * 86_400 + seconds;
+    let max_total_seconds = MAX_DAYS as i64 * 86_400;
+    if !(-max_total_seconds..=max_total_seconds).contains(&total_seconds) {
+        return write!(
+            w,
+            "{}",
+            ctx.failure(PlanError::invalid(
```
wackywendell: As above - push a warning, no need to `ctx.failure` here. The printer is supposed to be best-effort outputting, not strict - it should output anything it finds as long as it has a way to; failure tokens are only for cases where our syntax/grammar doesn't have a reasonable way to output it.

GRAMMAR.md
```diff
@@ -241,9 +241,44 @@ A literal can be an integer, float, boolean, string, or null. Literals may inclu
...
+
+#### `interval_day` Typed Literals
+
+`interval_day` string literals represent Substrait `IntervalDayToSecond` values.
+The string contains one to three whitespace-separated duration terms:
+
+```text
+interval_day_literal := duration_term (whitespace duration_term)? (whitespace duration_term)?
+duration_term        := "-"? digit+ duration_unit
+duration_unit        := "d" / "s" / "ms" / "us" / "ns" / "ps"
```
wackywendell: As above - I think it makes sense to:

- only allow a single space, not multiple, not tabs. We accept more in the broader grammar, but in literals, no need to.
- Only allow descending order, and only one subsecond unit.
- You can encode that into the PEG grammar here, too.

GRAMMAR.md
```diff
@@ -282,7 +317,13 @@ From [official Substrait grammar](https://raw.githubusercontent.com/substrait-io
 - `fp32`, `fp64`
 - `string`, `binary`
 - `timestamp`, `timestamp_tz`, `date`, `time`
-- `interval_year`, `uuid`
+- `interval_year`, `interval_day`, `uuid`
+
+Bare `interval_day` is accepted as a type name, with an unset precision
```
wackywendell: I don't think we should support this. `precision` is a required field ([now](https://github.com/substrait-io/substrait/blob/4bd6bcc437aea6fc0f75fc5be293ca2f86d22535/proto/substrait/algebra.proto#L1180) and [here](https://github.com/substrait-io/substrait/blob/4bd6bcc437aea6fc0f75fc5be293ca2f86d22535/proto/substrait/type.proto#L130)), let's require the user to be explicit.

src/parser/expressions.rs
```diff
@@ -363,6 +380,190 @@ fn parse_timestamp_to_microseconds(
...
+        return Err(MessageParseError::invalid(
+            "duration_parse_format",
+            span,
+            format!(
+                "Invalid duration '{duration_str}': subseconds ({subseconds}) must be less than one second at precision {precision} (magnitude < {subsecond_bound}); put whole seconds in the 's' term"
+            ),
+        ));
+    }
+
+    // Per the Substrait spec, IntervalDayToSecond supports a range of [-3,650,000..3,650,000] days.
```
wackywendell: Generally, we don't want to be stricter than necessary on Substrait - this isn't a validator, its a converter.

Generally - if 'invalid' Substrait has a clear syntax (as out-of-range seconds does), there's no need to go out of our way to reject it. Ideally, we would output a warning (as `textify` can), but in the parser, I'd suggest a `// TODO` note here about validation and just accept it.

wackywendell (changes requested):

src/parser/expressions.rs
gord02: So this is how PrecisionTimestamps is implemented and Claude helped derive these three reasons:
1. parse_pair returns Self, not Result — errors have nowhere to go, so failure can only panic. integer is "-"? ~ ASCII_DIGIT+ (unbounded), so '99999999999d' overflows days: i32; 

2. No channel for the precision — it comes from the type ascription (interval_day<9>), outside the duration string. 

3. No channel for the span. The string is unescaped into a temporary and re-parsed, so the second pass's spans point into memory, not the document. For '5d  3s':interval_day<0> we emit line 3, col 13 (the literal in context) by passing the outer span in; the inner parse's own span says 1:4 of a six-character temporary. It can't be fixed with an offset — unescape_string collapses \X → X, so '12\3ns' is 8 chars and 123ns is 5. parse_pair takes only the pair, and that pair's span is the inner one.

Also this bug is raised as well: The grammar's integer rule matches unbounded digits, but the parser handles overflow two different wrong ways. Digits that exceed the parse target panic — Project[$99999999999] hits .parse::<i32>().unwrap() at expressions.rs:62 — and values that fit i64 but not the literal's declared width are silently truncated by an cast, so 4294967296:i32 parses to 0:i32. Both should be MessageParseErrors; today one crashes the process and the other corrupts data without warning.

wackywendell: Ah, yes, passing in the precision would make sense (and you can add that here; it doesn't seem to be here?). Other than that - you shouldn't need to pass in the span; just return an error fitting what you find, and let the next layer up convert that to a `MessageParseError` with the span. (Don't be afraid to make a new intermediate error type; more types often make for more flexible code than code that reuses the same types for different things.)

I'd suggest splitting this: an outer function unescapes the string, calls `ExpressionParser::parse(Rule::interval_day_duration, unescaped_string)`, and handles the span and error creation. Then a second function (or `ParsePair`/`ScopedParsePair` implementation) that takes the resulting `Pair` and produces an `IntervalDayToSecond`.

gord02: Split as suggested — the outer/inner shape is there, just with the unescape one layer up. parse_interval_day_duration is the outer function: it calls ExpressionParser::parse(Rule::interval_day_duration, ..) and owns all error creation and span handling — both pest's syntax error and the inner error become a MessageParseError with the literal's span. interval_day_from_pair(pair, precision) takes the resulting Pair and produces the IntervalDayToSecond, returning a new IntervalDayError (TermOverflow, UnitPrecisionMismatch) that carries no span. Precision is passed to both, as SupportedPrecision.

The outer function was made to not unescape. This is because to_string_literal unescapes once for every typed string literal before dispatching on the type kind, so date, timestamp, and the rest already share that step. Moving it into the interval_day arm would mean unescaping twice or special-casing this one type in the dispatch. That's also why the span is an argument — the outer function has the unescaped String, not the Pair it came from, so it has nothing to derive a span from. Easy to change if you'd rather pass the string_literal pair down.
Neither ParsePair nor ScopedParsePair fit the inner half — the first returns Self with nowhere to put an error, the second is scoped by &SimpleExtensions rather than a precision — so it's a plain function.

src/parser/expression_grammar.pest
```diff
@@ -57,6 +57,33 @@ reference = { "$" ~ integer }
 // Literal
 literal = { (float | integer | boolean | string_literal | null) ~ (":" ~ sp ~ type)? }
 
+// -- interval_day durations --
```
wackywendell: This section should probably go at the end of the file, or even in its own separate file, since its not really part of the same "grammar" in the sense that its parsed separately...

src/parser/expressions.rs
```diff
@@ -449,6 +461,15 @@ fn duration_to_precision_units(
...
         // no finer fractional component that could be silently dropped here.
         SupportedPrecision::Nanoseconds => duration.num_nanoseconds().ok_or_else(out_of_range),
+        // `check_supported_precision` rejects picoseconds before we get here,
+        // since chrono has no sub-nanosecond resolution to convert into.
+        SupportedPrecision::Picoseconds => Err(MessageParseError::invalid(
+            "precision_literal_unsupported_precision",
+            span,
+            format!(
+                "precision 12 (picoseconds) is not supported for {literal_kind} literals; chrono only supports nanosecond (precision 9) resolution"
+            ),
```
wackywendell, line 461: 1. `unimplemented` is more accurate here than `unsupported` - we would like to implement it, just haven't yet. "Unsupported" makes it sound like we intentionally do not support it.
2. Details about the use of internal dependencies like `chrono` should not leak out in error messages; the error message should focus on what the user needs to know

```suggestion
        SupportedPrecision::Picoseconds => Err(MessageParseError::invalid(
            "precision_literal_unimplemented_precision",
            span,
            format!(
                "precision 12 (picoseconds) is not yet implemented for {literal_kind} literals"
            ),
```

src/parser/expressions.rs
```diff
@@ -457,29 +478,25 @@ fn check_supported_precision(
...
-        return Err(MessageParseError::invalid(
+    match SupportedPrecision::from_units(precision) {
+        // chrono has no sub-nanosecond resolution, so the literals that go
+        // through it can't represent picoseconds.
+        Some(SupportedPrecision::Picoseconds) => Err(MessageParseError::invalid(
             "precision_literal_unsupported_precision",
             span,
             format!(
                 "precision 12 (picoseconds) is not supported for {literal_kind} literals; chrono only supports nanosecond (precision 9) resolution"
             ),
```
wackywendell, line 479: As above. The comment here is good - code comments are where implementation details should be discussed; the error message should be at the user level of description. (I know this was already in here, but would make sense to fix).

```suggestion
"precision_literal_unimplemented_precision",
span,
format!(
"precision 12 (picoseconds) is not yet implemented for {literal_kind} literals"
),
```

src/parser/expressions.rs
```diff
@@ -547,6 +564,151 @@ fn parse_time_to_precision_units(
...
+    precision: SupportedPrecision,
+    span: pest::Span,
+) -> Result<IntervalDayToSecond, MessageParseError> {
+    let mut pairs = ExpressionParser::parse(Rule::interval_day_duration, duration_str).map_err(
+        |e| {
+            MessageParseError::invalid(
+                "interval_day_duration",
+                span,
+                format!(
+                    "Invalid duration '{duration_str}': {}. Expected one to three terms, separated by single spaces, in the order days ('d'), seconds ('s'), sub-seconds ('ms', 'us', 'ns', or 'ps'); e.g. '5d', '4d 5s', '5d 3s 100ns'",
```
wackywendell, line 622: Good error message 👍

src/textify/expressions.rs
```diff
@@ -188,6 +190,118 @@ fn precision_time_to_string(value: i64, precision: i32) -> Result<String, Precis
...
 
+/// The precision and sub-second count of an `IntervalDayToSecond`, as stored.
+///
+/// The deprecated `microseconds` precision mode is normalized to precision 6,
+/// since that field is microseconds by definition. An unset `precision_mode`
+/// also means microseconds: that is Substrait's documented meaning for interval
+/// values recorded before `precision_mode` existed, and it matches the default
+/// the type textifier uses for `Type.IntervalDay` with no precision.
+fn interval_day_precision_units(
+    interval: &substrait::proto::expression::literal::IntervalDayToSecond,
```
wackywendell: This should be `use` statement at the top, not a long inline like this
```suggestion
    interval: &IntervalDayToSecond,
```

src/precision.rs
```diff
@@ -0,0 +1,105 @@
+//! Sub-second precision for literal values, shared by the parser and textifier.
```
wackywendell, line 1: 👍 , good module. 🤔  Organizationally, it might make sense to find a home for this in a directory in `src` rather than top-level like this, but its fine for now - and good that its all in one file 👍

wackywendell (approved):
This is looking much better! The split between `parse_interval_day_duration` and `interval_day_from_pair` makes sense.

I have a couple code organization comments, but otherwise, LGTM!

PR #185, opened 2026-07-15
feat: implementing window function expressions support

### LINKED ISSUE #158: [FEATURE] Support for WindowFunction
### Background

Expression.WindowFunction is currently unimplemented.

#### Proto fields

Shared with ScalarFunction/AggregateFunction: function_reference, arguments, options, output_type

Window-specific required: phase (AggregationPhase), bounds_type (BoundsType)

Window-specific optional: partitions, sorts, invocation, lower_bound, upper_bound

Note: AggregateFunction already shares `phase`/`invocation` with WindowFunction, but currently drops both on textify (`src/parser/expressions.rs:680-681`, marked TODO) — there is no existing precedent for how these two fields should be displayed.

### Option 1 — inline function call with trailing named args

Window functions are valid inline expressions, displayed like scalar/aggregate function calls (`name(args):type`) with window-specific fields appended as trailing named arguments inside the same parens. Enum-valued fields (`phase`, `invocation`, `bounds`) reuse the existing `&Name` `enum_value` token (same as `join_type`/`sort_direction`); `order=` reuses the existing `sort_field` production verbatim; named-arg `key=value` syntax matches the existing `Fetch[limit=..., offset=...]` and extension-relation named-argument conventions. Only non-default fields need be emitted.

```
sum($0, partition=($1, $2), order=(($3, &AscNullsLast)), phase=&InitialToResult, invocation=&Distinct, bounds=&Rows, lower=-3, upper=current):fp64
row_number(phase=&InitialToResult):i64
```

```
window_function_call = {
    compound_name ~ sp ~ anchor? ~ sp ~ urn_anchor? ~ sp
    ~ "(" ~ sp ~ argument_expr_list? ~ (sp ~ "," ~ sp ~ window_named_arg)* ~ sp ~ ")"
    ~ ":" ~ sp ~ type
}

argument_expr_list = { expression ~ (sp ~ "," ~ sp ~ expression)* }

window_named_arg = {
    ("phase"      ~ sp ~ "=" ~ sp ~ enum_value) |
    ("invocation" ~ sp ~ "=" ~ sp ~ enum_value) |
    ("bounds"     ~ sp ~ "=" ~ sp ~ enum_value) |
    ("lower"      ~ sp ~ "=" ~ sp ~ window_bound) |
    ("upper"      ~ sp ~ "=" ~ sp ~ window_bound) |
    ("partition"  ~ sp ~ "=" ~ sp ~ "(" ~ sp ~ expression ~ (sp ~ "," ~ sp ~ expression)* ~ sp ~ ")") |
    ("order"      ~ sp ~ "=" ~ sp ~ "(" ~ sp ~ sort_field  ~ (sp ~ "," ~ sp ~ sort_field)*  ~ sp ~ ")")
}

window_bound = { "unbounded" | "current" | ("-" ~ integer) | ("+" ~ integer) }
```

Valid anywhere an `expression` is valid, satisfying the proto's `Expression.RexType` requirement directly — no structural restriction on where a window function can appear.

### Option 2 — multiline named-binding block

Window functions are displayed as named output bindings on their own indented lines within a relation, with all window-specific fields as named arguments (reusing `window_named_arg` from Option 1 verbatim) inside a `[...]` clause after the argument list. `$a:` assigns the result to an output column reference. 

```
Project[$0, $2, $a]
  $a: sum($1)[
    partition=($2, $4), order=(($3, &AscNullsLast), ($5, &DescNullsFirst)), phase=&InitialToResult, invocation=&Distinct, bounds=&Rows, lower=-3, upper=current
  ]
```

```
window_binding = {
    "$" ~ identifier ~ ":" ~ sp ~ compound_name ~ sp ~ anchor? ~ sp ~ urn_anchor? ~ sp
    ~ argument_list
    ~ sp ~ "[" ~ sp ~ window_named_arg ~ (sp ~ "," ~ sp ~ window_named_arg)* ~ sp ~ "]"
}
```
*(`window_named_arg`, `window_bound` as defined in Option 1.)*

### Option 3 — multiline named-binding block with `window` keyword prefix

Identical to Option 2, but adds a bare leading keyword before the function name to make the construct self-identifying at a glance, matching how relations already self-identify with a leading bare name (`Project`, `Filter`, `Read:Virtual`). 
```
Project[$0, $2, $a]
  $a: window sum($1)[
    partition=($2, $4),
    order=(($3, &AscNullsLast), ($5, &DescNullsFirst)),
    phase=&InitialToResult,
    invocation=&Distinct,
    bounds=&Rows,
    lower=-3,
    upper=current
  ]
```

```
window_binding = {
    "$" ~ identifier ~ ":" ~ sp ~ "window" ~ sp ~ compound_name ~ sp ~ anchor? ~ sp ~ urn_anchor? ~ sp
    ~ argument_list
    ~ sp ~ "[" ~ sp ~ window_named_arg ~ (sp ~ "," ~ sp ~ window_named_arg)* ~ sp ~ "]"
}
```
*(`window_named_arg`, `window_bound` as defined in Option 1.)*

gord02:
## Description

Adds parsing and textifying for Substrait's `Expression.WindowFunction`, following the existing scalar/aggregate function syntax with an `over(...)` clause for window-specific args.

**Syntax:** `func(args) over(phase=..., partition=(...), order=..., invocation=..., rows=(lo, hi) | range=(lo, hi)):type`

- `phase=` (required) and `invocation=` map to `AggregationPhase`/`AggregationInvocation`.
- `partition=(...)` — partitioning expressions.
- `order=` — sort field(s); bare for a single field, parenthesized list of tuples for multiple, matching `Sort`'s `sort_field_list` convention.
- `rows=(lo, hi)` / `range=(lo, hi)` — frame bounds, where each side is an integer offset, `0` for current row, or `_` for unbounded (**parsed to an explicit `Bound::Unbounded`, not left as `None`**).
- `range=` is rejected when more than one `order=` field is present, matching the proto's semantics.

**Changes:**
- `src/parser/expression_grammar.pest` — new grammar rules for `window_function_call` and its `over(...)` args.
- `src/parser/expressions.rs` — `WindowFunction::parse_pair`.
- `src/textify/expressions.rs` — corresponding textifier.
- `GRAMMAR.md` — documents the new syntax.
- `tests/plan_roundtrip.rs` — round-trip integration tests, alongside new unit tests in the parser/textify modules.

## Type of Change
- [x] New feature

## Testing
- [x] All existing tests pass

## Related Issues
Closes #158

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -507,28 +507,26 @@ A window function computes a value over a "window" of rows related to the curren
 
 #### Syntax
 
-`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" (window_named_arg ("," window_named_arg)*)? ")" ":" type`
+`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" window_named_arguments? ")" ":" type`
 
-```text
-window_named_arg := "partition=" "(" expression ("," expression)* ")"
-                   / "order=" sort_field
-                   / "order=" "(" sort_field ("," sort_field)+ ")"
-                   / "invocation=" enum
-                   / "phase=" enum
-                   / ("rows=" / "range=") "(" window_bound "," window_bound ")"
-window_bound := integer / "_"
-```
+Rather than a bespoke grammar production per named argument, `over(...)` reuses the same
+generic [`named_arguments`](#arguments) production used elsewhere (`name "=" argument`,
+comma-separated). Rust code (not the grammar) decides which names are allowed in `over(...)`
+and what shape each value must have:
 
-`sort_field` is the same production used by the Sort relation (see below): `"(" reference "," sort_direction ")"`.
+- `partition=argument` - optional partitioning expression(s); a single field may be written
+  bare (`partition=$1`) or in a one-element tuple (`partition=($1,)`); multiple fields use a
+  multi-element tuple (`partition=($1, $2)`)
+- `order=argument` - optional sort key(s); a single sort field is a bare 2-tuple of
+  `(reference, direction)` (e.g. `order=($3, &AscNullsLast)`); multiple sort fields are a
+  tuple of such 2-tuples (e.g. `order=(($3, &AscNullsLast), ($4, &DescNullsLast))`)
+- `invocation=&Distinct` / `invocation=&All` - optional aggregation invocation
+- `phase=&InitialToResult` (etc.) - required aggregation phase
+- `rows=(lower, upper)` / `range=(lower, upper)` - optional, mutually exclusive window frame;
+  each bound is an integer (negative = preceding, positive = following, `0` = current row) or
+  `_` for unbounded
 
-- `partition=(expression, ...)` - partitioning expressions
-- `order=sort_field` or `order=(sort_field, sort_field, ...)` - ordering field(s); a single sort field is written bare (e.g. `order=($1, &AscNullsLast)`), two or more are wrapped in a parenthesized list of tuples (e.g. `order=(($1, &AscNullsLast), ($2, &DescNullsLast))`)
-- `invocation=&Distinct` / `invocation=&All` - aggregation invocation
-- `phase=&InitialToResult` (etc.) - aggregation phase
-- `rows=(lower, upper)` / `range=(lower, upper)` - the window frame, each bound is an integer (negative = preceding,
-  positive = following, `0` = current row) or `_` for unbounded
-
-`partition=`, `order=`, `invocation=`, and `rows=`/`range=` are all optional and are omitted when empty; `phase=` is always required and always printed, though this is enforced when parsing rather than by the grammar itself (the named-arg list as a whole is syntactically optional). A `range=` frame requires exactly one `order=` field.
+`partition=`, `order=`, `invocation=`, and `rows=`/`range=` are all optional and are omitted when empty; `phase=` is always required and always printed, though this (along with the mutual exclusion of `rows=`/`range=`, and the requirement that a `range=` frame have exactly one `order=` field) is enforced when parsing rather than by the grammar itself (the named-arg list as a whole is syntactically optional).
 
 #### Examples
 
@@ -544,7 +542,7 @@ Functions:
 
 === Plan
 Root[a, b, s]
-  Project[$0, $1, sum($1) over(phase=&InitialToResult, order=($1, &AscNullsLast), invocation=&Distinct, partition=($0), rows=(-3, 0)):fp64?]
+  Project[$0, $1, sum($1) over(phase=&InitialToResult, order=($1, &AscNullsLast), invocation=&Distinct, partition=$0, rows=(-3, 0)):fp64?]
     Read[t => a:i32, b:fp64]
 # "#;
 #
@@ -566,7 +564,7 @@ Functions:
 
 === Plan
 Root[a, r]
-  Project[$0, row_number() over(phase=&InitialToResult, order=($0, &AscNullsLast), partition=($0)):i64]
+  Project[$0, row_number() over(phase=&InitialToResult, order=($0, &AscNullsLast), partition=$0):i64]
     Read[t => a:i32]
 # "#;
 #
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -509,10 +509,19 @@ A window function computes a value over a "window" of rows related to the curren
 
 `window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" (window_named_arg ("," window_named_arg)*)? ")" ":" type`
 
-`window_named_arg` is one of:
+```text
+window_named_arg := "partition=" "(" expression ("," expression)* ")"
+                   / "order=" sort_field
+                   / "order=" "(" sort_field ("," sort_field)+ ")"
+                   / "invocation=" enum
+                   / "phase=" enum
+                   / ("rows=" / "range=") "(" window_bound "," window_bound ")"
+sort_field := "(" reference "," enum ")"
+window_bound := integer / "_"
+```
 
 - `partition=(expression, ...)` - partitioning expressions
-- `order=sort_field` or `order=(sort_field, sort_field, ...)` - ordering field(s); a single sort field is written bare (e.g. `order=($1,&AscNullsLast)`), two or more are wrapped in a parenthesized list of tuples (e.g. `order=(($1,&AscNullsLast),($2,&DescNullsLast))`)
+- `order=sort_field` or `order=(sort_field, sort_field, ...)` - ordering field(s); a single sort field is written bare (e.g. `order=($1, &AscNullsLast)`), two or more are wrapped in a parenthesized list of tuples (e.g. `order=(($1, &AscNullsLast), ($2, &DescNullsLast))`)
 - `invocation=&Distinct` / `invocation=&All` - aggregation invocation
 - `phase=&InitialToResult` (etc.) - aggregation phase
 - `rows=(lower, upper)` / `range=(lower, upper)` - the window frame, each bound is an integer (negative = preceding,
@@ -534,7 +543,7 @@ Functions:
 
 === Plan
 Root[a, b, s]
-  Project[$0, $1, sum($1) over(phase=&InitialToResult, order=($1,&AscNullsLast), invocation=&Distinct, partition=($0), rows=(-3, 0)):fp64?]
+  Project[$0, $1, sum($1) over(phase=&InitialToResult, order=($1, &AscNullsLast), invocation=&Distinct, partition=($0), rows=(-3, 0)):fp64?]
     Read[t => a:i32, b:fp64]
 # "#;
 #
@@ -556,7 +565,7 @@ Functions:
 
 === Plan
 Root[a, r]
-  Project[$0, row_number() over(phase=&InitialToResult, order=($0,&AscNullsLast), partition=($0)):i64]
+  Project[$0, row_number() over(phase=&InitialToResult, order=($0, &AscNullsLast), partition=($0)):i64]
     Read[t => a:i32]
 # "#;
 #
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -516,9 +516,10 @@ window_named_arg := "partition=" "(" expression ("," expression)* ")"
                    / "invocation=" enum
                    / "phase=" enum
                    / ("rows=" / "range=") "(" window_bound "," window_bound ")"
-sort_field := "(" reference "," enum ")"
 window_bound := integer / "_"
 ```
+
+`sort_field` is the same production used by the Sort relation (see below): `"(" reference "," sort_direction ")"`.
 
 - `partition=(expression, ...)` - partitioning expressions
 - `order=sort_field` or `order=(sort_field, sort_field, ...)` - ordering field(s); a single sort field is written bare (e.g. `order=($1, &AscNullsLast)`), two or more are wrapped in a parenthesized list of tuples (e.g. `order=(($1, &AscNullsLast), ($2, &DescNullsLast))`)
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -507,7 +507,7 @@ A window function computes a value over a "window" of rows related to the curren
 
 #### Syntax
 
-`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" window_named_arguments? ")" ":" type`
+`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" named_arguments ")" ":" type`
 
 Rather than a bespoke grammar production per named argument, `over(...)` reuses the same
 generic [`named_arguments`](#arguments) production used elsewhere (`name "=" argument`,
@@ -526,7 +526,7 @@ and what shape each value must have:
   each bound is an integer (negative = preceding, positive = following, `0` = current row) or
   `_` for unbounded
 
-`partition=`, `order=`, `invocation=`, and `rows=`/`range=` are all optional and are omitted when empty; `phase=` is always required and always printed, though this (along with the mutual exclusion of `rows=`/`range=`, and the requirement that a `range=` frame have exactly one `order=` field) is enforced when parsing rather than by the grammar itself (the named-arg list as a whole is syntactically optional).
+`partition=`, `order=`, `invocation=`, and `rows=`/`range=` are all optional and are omitted when empty; `phase=` is always required and always printed. The grammar itself only requires that at least one named argument be present; the specific rules — that `phase=` must appear, that `rows=`/`range=` are mutually exclusive, and that a `range=` frame have exactly one `order=` field — are all enforced by the parser rather than by the grammar.
 
 #### Examples
 
@@ -616,13 +616,17 @@ Arguments in relations can be literals, expressions, enums, or tuples thereof.
 #### Syntax
 
 ```text
-argument := enum / reference / literal / expression / tuple
+argument := enum / reference / literal / expression / tuple / "_"
 tuple := "(" ")"                                        // 0-tuple
        / "(" argument "," ")"                           // 1-tuple (trailing comma required)
        / "(" argument ("," argument)+ ","? ")"          // 2+-tuple (trailing comma optional)
 arguments := argument ("," argument)*
 named_arguments := name "=" argument ("," name "=" argument)*
 ```
+
+`_` is accepted syntactically as an argument (it denotes an unbounded window frame
+bound in `over(...)`), but is only meaningful there; consumers that don't support it —
+extension-relation arguments, for example — reject it during parsing.
 
 Tuples follow the Python/Rust trailing-comma convention to disambiguate from parenthesised
 expressions: `(x)` is a parenthesised expression, not a tuple. A trailing comma is required to
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -507,7 +507,7 @@ A window function computes a value over a "window" of rows related to the curren
 
 #### Syntax
 
-`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" window_named_arg ("," window_named_arg)* ")" ":" type`
+`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" (window_named_arg ("," window_named_arg)*)? ")" ":" type`
 
 `window_named_arg` is one of:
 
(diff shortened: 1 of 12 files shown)


src/parser/expressions.rs
```diff
@@ -1450,4 +1678,110 @@ mod tests {
             "u! prefix in function call base name must be rejected by the grammar"
         );
     }
+
+    #[test]
+    fn test_window_function_full() {
+        let exts = make_extensions_for_fn_tests();
```
gord02, line 1976: nit: change this name

src/parser/expressions.rs
```diff
@@ -458,6 +462,225 @@ impl ScopedParsePair for ScalarFunction {
...
+        let mut invocation = AggregationInvocation::Unspecified as i32;
+        let mut phase = None;
+        let mut bounds_type = window_function::BoundsType::Unspecified as i32;
+        let mut lower_bound = None;
+        let mut upper_bound = None;
+
+        for arg in named_arg_list.into_inner() {
+            assert_eq!(arg.as_rule(), Rule::window_named_arg);
+            let inner = unwrap_single_pair(arg);
+            match inner.as_rule() {
```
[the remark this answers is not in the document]
gord02: is this case worth addressing?

src/parser/expression_grammar.pest
```diff
@@ -179,10 +183,43 @@ cast_failure_behavior = { "?" | "!" }
...
+window_partition_arg  = { "partition" ~ sp ~ "=" ~ sp ~ "(" ~ sp ~ expression_list? ~ sp ~ ")" }
+window_sort_order      = { sort_field | ("(" ~ sp ~ sort_field ~ (sp ~ "," ~ sp ~ sort_field)+ ~ sp ~ ")") }
+window_order_arg       = { "order" ~ sp ~ "=" ~ sp ~ window_sort_order }
+window_bound           = { integer | empty }
+window_frame_kind      = { "rows" | "range" }
+window_frame_arg       = { window_frame_kind ~ sp ~ "=" ~ sp ~ "(" ~ sp ~ window_bound ~ sp ~ "," ~ sp ~ window_bound ~ sp ~ ")" }
+window_phase_arg       = { "phase" ~ sp ~ "=" ~ sp ~ enum_value }
+window_invocation_arg  = { "invocation" ~ sp ~ "=" ~ sp ~ enum_value }
+
+window_named_arg      = { window_partition_arg | window_order_arg | window_frame_arg | window_phase_arg | window_invocation_arg }
```
wackywendell: Could we reuse the common argument value, tuple, and named-argument grammar here, rather than define a separate production for each window field? The current approach gives window functions a parallel argument syntax, even though Rust still needs to interpret the fields afterwards.

I would expect Pest to parse the same generic values we use elsewhere, with the window parser then deciding which names are allowed and what shape each value must have. That would also give us one place in Rust to reject duplicate fields, require `phase`, and make `rows` and `range` mutually exclusive.

Tuple syntax should work the same way here as it does elsewhere: `(x,)` is a one-element tuple and `(x, y)` is a two-element tuple. Where a window field represents a list, its conversion can accept either one element or a tuple. For example, `order=($0, &AscNullsLast)` is one sort field, while `order=(($0, &AscNullsLast),)` is an explicit one-element tuple of sort fields. Canonical output can continue to use the shorter form.

src/parser/expressions.rs
```diff
@@ -396,6 +401,47 @@ impl ScopedParsePair for Literal {
     }
 }
 
+/// The shared prefix of function calls.
+struct FunctionHead {
```
wackywendell: I'm not sure `FunctionHead` is the right abstraction here. It includes the function arguments, but leaves the output type and window clause for the caller to parse. Its name describes where it appears in the input rather than a complete thing in the syntax, so callers still need to know which pieces remain in the iterator.

Could we instead parse an unresolved function reference, arguments, an over clause, and the output type as meaningful components? Those could compose into complete `ScalarFunctionInvocation` and `WindowFunctionInvocation` syntax types. Conversion from those types could then resolve the function reference and construct the protobuf message.

src/parser/expressions.rs
```diff
@@ -396,6 +401,47 @@ impl ScopedParsePair for Literal {
...
+    name: CompoundName,
+    anchor: Option<u32>,
+    arguments: Vec<FunctionArgument>,
+}
+
+/// Parse the `CompoundName ~ anchor? ~ urn_anchor? ~ argument_list` prefix of functions.
+fn parse_function_head(
```
wackywendell: Could the syntax types above implement the existing pair-parsing traits? Right now `parse_function_head` is a standalone function, even though `ParsePair` and `ScopedParsePair` normally connect a grammar rule directly to the Rust type it produces.

Function arguments need extension context, so `ScopedParsePair` seems like the natural fit for the invocation types. That would keep the rule and its parsing behavior on the parsed type, instead of asking callers to coordinate an iterator with a separate helper.

src/textify/expressions.rs
```diff
@@ -1,19 +1,25 @@
...
 use substrait::proto::function_argument::ArgType;
 use substrait::proto::{
-    AggregateFunction, Expression, FunctionArgument, FunctionOption, expression as expr,
+    AggregateFunction, AggregationPhase, Expression, FunctionArgument, FunctionOption, SortField,
+    expression as expr,
 };
 
 use super::{PlanError, Scope, Textify, Visibility};
 use crate::extensions::simple::ExtensionKind;
+use crate::textify::rels::{Arguments, NamedArg, Value, ValueEnum, enum_str_value};
```
wackywendell: Could we avoid making expression formatting depend on the relation formatter? `Arguments`, `NamedArg`, `Value`, and `ValueEnum` describe shared text-format concepts, but they currently live in `textify::rels`. This change also puts the window-function enum implementations there, which makes it less clear which module owns this behavior.

One option would be to move `Value`, `NamedArg`, and the related enum conversion into a neutral textify module; you could also (instead?) move window functions into their own module. I'm open to other arrangements, but let's try and make them fairly well-defined/separate so that relations can depend on expressions (and treat it as a 'black box'), and not vice versa.

src/textify/expressions.rs
```diff
@@ -522,6 +541,158 @@ impl Textify for IfThen {
     }
 }
 
+fn window_enum_value<'a, T: TryFrom<i32> + ValueEnum>(
```
wackywendell: This interface looks like a symptom of the module boundary above. `window_enum_value` is generic over the protobuf enum type, but hard-codes `WindowFunction` in its errors. It also combines protobuf decoding, diagnostics, and formatting, and its return lifetime is unrelated to any input because `ValueEnum` only provides a borrowed conversion.

Could we give decoded enum values a direct, owned conversion into the shared formatting value instead? The window textifier could then keep responsibility for decoding its raw fields and reporting field-specific errors. I would prefer that over renaming or further generalizing this helper.

src/textify/expressions.rs
```diff
@@ -522,6 +541,158 @@ impl Textify for IfThen {
...
+}
+
+fn window_sort_order<'a>(sorts: &'a [SortField]) -> Value<'a> {
+    match sorts {
+        [single] => Value::from(single),
+        many => Value::Tuple(many.iter().map(Value::from).collect()),
+    }
+}
+
+fn window_bound_value<'a>(
```
wackywendell, line 675: This can silently turn malformed protobuf bounds into different valid variants. A `Preceding` or `Following` offset of zero is written as `0`, which parses back as `CurrentRow`. Also, `bound.and_then(...)` treats both `None` and `Some(Bound { kind: None })` as `_`, so the malformed value parses back as `Unbounded`.

Could we distinguish `None`, which is a valid default, from `Some(Bound { kind: None })`, which is malformed? The latter should produce a diagnostic rather than textifying as unbounded. `Preceding` and `Following` also require strictly positive offsets, so zero should produce a diagnostic rather than textifying as `CurrentRow`. Could we add focused tests for those two cases?

src/parser/expression_grammar.pest
```diff
@@ -179,10 +183,43 @@ cast_failure_behavior = { "?" | "!" }
...
+    function_call_prefix ~ sp ~ "over" ~ sp ~ "(" ~ sp ~ window_named_arg_list ~ sp ~ ")" ~ ":" ~ sp ~ type
+}
+
 // Top-level Expression Rule
 // Order matters for PEGs: Since an identifier can be a function call, we put that first.
 // cast_expression must come before literal/function_call since it starts with "("
-expression = { if_then | cast_expression | function_call | reference | literal }
+// window_function_call must come before function_call: both share the same
+// prefix, but window_function_call requires the trailing "over(...)" clause.
+expression = { if_then | cast_expression | window_function_call | function_call | reference | literal }
```
wackywendell, line 213: 👍

src/parser/expressions.rs
```diff
@@ -458,6 +488,208 @@ impl ScopedParsePair for ScalarFunction {
     }
 }
 
+/// Parse a `window_bound` pair (`integer | empty`) into a `window_function::Bound`.
+/// `_` (empty) means unbounded, represented explicitly as `Some(Bound { kind:
+/// Some(Unbounded{}) })` rather than `None`. `0` maps to `CurrentRow`;
+/// negative/positive integers map to `Preceding`/`Following`.
+fn parse_window_bound(
```
wackywendell: This is quite confusing; what does `None` mean? Do we ever return `None`, or should we just drop it?

Also, a good candidate for a trait implementation - `impl ParsePair for window_function::Bound`.

src/parser/expressions.rs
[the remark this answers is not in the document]
wackywendell: Yes - we should probably either catch these and error, or at least a TODO comment, rather than silently ignore.

src/textify/expressions.rs
```diff
@@ -402,32 +408,45 @@ impl Textify for FieldReference {
     }
 }
 
+/// Write the `name(args, options)` prefix shared by `ScalarFunction`,
+/// `AggregateFunction`, and `WindowFunction` textification.
+fn textify_function_call_prefix<S: Scope, W: fmt::Write>(
```
wackywendell, line 413: As with the parser - I think this makes more sense to break into components: function name/reference; arguments; over clause; output type. Implement `Textify` for each of these (perhaps by creating a newtype for each, as needed), then `Textify` for a FunctionCall and for a WindowFunctionCall can just make four calls to their elements.

tests/plan_roundtrip.rs
wackywendell, line 1: Good tests :+1:

GRAMMAR.md
```diff
@@ -501,6 +501,79 @@ Root[status]
 #  assert_eq!(plan.relations.len(), 1);
 ```
 
+### Window Functions
+
+A window function computes a value over a "window" of rows related to the current row (partitioning, ordering, and an optional frame), rather than collapsing rows the way an aggregate does.
+
+#### Syntax
+
+`window_function := function_signature anchor? urn_anchor? "(" (expression ("," expression)*)? ")" "over(" (window_named_arg ("," window_named_arg)*)? ")" ":" type`
```
wackywendell: As in the Pest grammar, I would define the arguments in `over(…)` as `named_arguments` here, which are already defined.

Below, defining the specific arguments allowed in terms of a "signature" for each (e.g. `partition` is an expression, or tuple of expressions), using terms as defined above, would be good.

src/textify/expressions.rs
```diff
@@ -1135,4 +1289,255 @@ mod tests {
...
+            .with_function(1, 10, "row_number");
+        let f = base_window_function();
+        assert_eq!(
+            ctx.textify_no_errors(&f),
+            "row_number() over(phase=&InitialToResult):i16"
+        );
+    }
+
+    #[test]
+    fn test_window_function_full() {
```
wackywendell, line 1388: There are a lot of complicated tests here, and by their nature, they are a bit hard to follow; do we need all of them, or are they covered by the roundtrip tests?

wackywendell (changes requested):
This is a good start for window-function support, especially the round-trip coverage. I think the parser will be easier to extend if window functions reuse the existing argument and tuple syntax, with Rust responsible for interpreting the named fields. On the formatting side, the `expressions` module should not depend on the `rels` module; it should only go the other way to make the separation of concerns clearer, and we should reorganize as needed to achieve that.

There are also a couple of conversion cases to tighten up: representable range frames should not be rejected during parsing, and malformed protobuf bounds should produce diagnostics rather than silently changing variants. I'd like another pass on those boundaries before approval, but the overall direction looks good.

gord02:
A lot has changed in this last review so I tried to summarize the changes claude made:

---

## Grammar: reuse the generic argument grammar for `over(...)`
**Resolution:** Deleted the five window-specific productions (`window_value`/`window_tuple`/`window_named_arg`/`window_named_arg_list`). `over(...)` now uses the generic `extension_argument` / `tuple` / `extension_named_arguments` rules. To let `_` (unbounded bound) be a tuple element, `empty` was added to `extension_argument`; `ExtensionValue::parse_pair` (`src/parser/extensions.rs`) now rejects `_` with a typed `MessageParseError`. All shape/semantic interpretation lives in Rust (`OverClause`), which is the single place that rejects duplicates/unknowns (`ParsedNamedArgs`), requires `phase=`, and enforces `rows`/`range` exclusivity.

- `src/parser/expression_grammar.pest`, `src/parser/extensions.rs`, `src/parser/expressions.rs` (`OverClause`)

---

## Confusing `Option<Bound>` return / use a trait

**Comment** — wackywendell, `expressions.rs:495` (`3605333063`): "what does `None` mean? … good candidate for `impl ParsePair for window_function::Bound`."

**Resolution:** Replaced the standalone `parse_window_bound` (which never actually returned `None`) with `impl ScopedParsePair for window_function::Bound` returning
`Result<Bound, _>`. `ScopedParsePair` rather than `ParsePair` because the foreign type can't take an inherent impl (orphan rules) and the `checked_neg` path must be
fallible; the `_extensions` param mirrors the existing `SortField` precedent.
---

## Silently-ignored malformed window arg shapes

**Comment** — wackywendell, `expressions.rs:588` (`3605337174`): "catch these and error … rather than silently ignore."

**Resolution:** The new arg helpers (`parse_window_order`, `parse_window_frame`, `sort_field_from_tuple`) return `MessageParseError` on wrong shapes/counts (via
`<[_; 2]>::try_from`) instead of asserting/ignoring. Covered by the non-integer and wrong-arity cases in the rejection table.

- `src/parser/expressions.rs`

---

## `FunctionHead` abstraction / trait-based parsing

**Comments** — wackywendell, `expressions.rs:405` (`3604792068`) and `:412` (`3604792070`): `FunctionHead` names where it sits in the input and leaves pieces
in the caller's iterator; parse meaningful components as syntax types that implement the pair-parsing traits.

**Resolution:** Deleted `FunctionHead`/`parse_function_head`. Introduced `FunctionReference` (unresolved name+anchor, `impl ParsePair`, with a `resolve()`
that does the registry lookup) and a new named `function_reference` grammar rule. `WindowFunctionInvocation` + `OverClause` (`ScopedParsePair`) compose the pieces
and a `resolve()` builds the protobuf. `ScalarFunction::parse_pair` composes `FunctionReference` + `parse_argument_list` + type directly (a transient
`ScalarFunctionInvocation` layer was added then collapsed, per the follow-up review, to keep one-type-per-rule for the scalar path).

- `src/parser/expression_grammar.pest`, `src/parser/expressions.rs`

---

## Expression formatter depending on the relation formatter

**Comment** — wackywendell, `textify/expressions.rs:22` (`3604792073`): move `Value`/`NamedArg`/`Arguments`/`ValueEnum` out of `textify::rels` so relations can
depend on expressions and not vice versa.

**Resolution:** Extracted them (plus `ArgsLayout`, `enum_str_value`) into a new neutral `src/textify/values.rs`. Both `rels` and `expressions` now depend
downward on it. The one residual back-dependency (`values → expressions::Reference`) was removed by inlining `write!("${i}")`, so `values.rs` is a true leaf.

- `src/textify/values.rs` (new), `src/textify/rels.rs`, `src/textify/expressions.rs`, `src/textify/mod.rs`

---

## `window_enum_value` helper (symptom of the module boundary)

**Comment** — wackywendell, `textify/expressions.rs:544` (`3604792076`): generic helper hard-codes `WindowFunction` in errors and bundles decode+diagnostic+format;
give decoded enums a direct owned conversion into the shared value.

**Resolution:** Removed `window_enum_value`. The window textifier decodes its raw `i32` field via the shared `decode_enum_field::<T>()` (in `values.rs`), which
reports a field-specific error tagged with the enum's own message type (`AggregationPhase` / `AggregationInvocation`) and uses the owned `enum_str_value`
conversion. `decode_enum_field` is also applied to `SetRel` to remove the same decode duplication.

- `src/textify/values.rs`, `src/textify/expressions.rs`, `src/textify/rels.rs`

---

## Malformed bounds silently become valid variants

**Comments** — wackywendell `textify/expressions.rs:690` (`3604792080`) and chatgpt-codex `:710` (`3597639246`): `Preceding`/`Following` offset `0` renders as
`0` (re-parses as `CurrentRow`); `None` and `Some(Bound{kind:None})` both render as _`; reject zero offsets and distinguish the malformed case.

**Resolution:** `window_bound_value` now distinguishes an absent bound (`None` → `_`, valid) from a present-but-empty bound (`Some(kind:None)` → diagnostic), and
emits a diagnostic for `Preceding`/`Following` offset `0` instead of rendering it. Covered by `test_window_function_bound_with_no_kind_surfaces_error` and
`test_window_function_zero_offset_bound_surfaces_error`.

- `src/textify/expressions.rs`
---

## Split the window `Textify` impl into components

**Comment** — wackywendell, `textify/expressions.rs:413` (`3605352192`): break into name/reference, arguments, over-clause, output type.

**Resolution:** `impl Textify for WindowFunction` is now prefix → over-clause → output type, delegating the over-clause to `window_over_named_args` / `window_frame_named_arg` / `window_partition_value`.

- `src/textify/expressions.rs`

---

## GRAMMAR.md `over(...)` docs

**Comment** — wackywendell, `GRAMMAR.md:510` (`3605385079`): define `over(…)` args as the already-defined `named_arguments`, with a per-field signature.

**Resolution:** The window section now references the generic `named_arguments` production and documents each field's signature (`partition=`, `order=`,
`rows=`/`range=`, `phase=`, `invocation=`).

- `GRAMMAR.md`

---

## Test-related

**Comment** — wackywendell, `textify/expressions.rs:1403` (`3605406192`): "a lot of complicated tests … do we need all of them, or are they covered by the roundtrip
tests?"

**Resolution:** The textify-only tests are kept — they exercise malformed / hand- built protos (`kind:None`, zero offsets, unknown enum i32s, wrong sort counts)
that the parser cannot produce, so round-trip tests structurally can't reach them. Separately, the parser-side rejection tests were consolidated from many one-off
functions into a single table-driven `test_window_function_invalid_over_clauses_rejected` (matching the repo's existing `for input in …` test style).

- `src/parser/expressions.rs`, `tests/plan_roundtrip.rs`

---

## Known limitation, intentionally not fixed here

`SortDirection::Clustered` textifies as `&Clustered` but is not accepted by the `sort_direction` grammar rule / parser, so a Clustered sort does not round-trip.
This is **pre-existing on `main`** (affects the `Sort` relation independently of window functions); this PR neither introduces nor fixes it. Best addressed as a
separate change that also covers the `Sort` relation.

PR #186, opened 2026-07-21
refactor(textify): extract shared textify components

gord02:
## Description

This PR refactors the textify code for expressions into its own module file that will be used by textify/expressions.rs and textify/rels.rs.

### Key Changes
- Move Value, NamedArg, Arguments, ArgsLayout, ValueEnum, enum_str_value, and the SortField->Value conversion out of textify::rels into a new neutral textify::values module
- adds a shared decode_enum_field helper (dedups the raw-i32-enum -> Value decode, applied here to SetRel). This decouples textify::expressions from textify::rels: both now depend downward on textify::values

## Testing
- [x] All existing tests pass

### DIFF
--- a/src/textify/expressions.rs
+++ b/src/textify/expressions.rs
@@ -1,4 +1,4 @@
-use std::fmt::{self};
+use std::fmt;
 
 use chrono::{DateTime, NaiveDate};
 use expr::RexType;
@@ -402,32 +402,45 @@ impl Textify for FieldReference {
     }
 }
 
+/// Write the `name(args, options)` prefix shared by `ScalarFunction`,
+/// `AggregateFunction`, and `WindowFunction` textification.
+fn textify_function_call_prefix<S: Scope, W: fmt::Write>(
+    function_reference: u32,
+    arguments: &[FunctionArgument],
+    options: &[FunctionOption],
+    ctx: &S,
+    w: &mut W,
+) -> fmt::Result {
+    let name_and_anchor = NamedAnchor::lookup(ctx, ExtensionKind::Function, function_reference);
+    let name_and_anchor = ctx.display(&name_and_anchor);
+
+    let between = if arguments.is_empty() || options.is_empty() {
+        ""
+    } else {
+        ", "
+    };
+    let args = ctx.separated(arguments, ", ");
+    let options = ctx.separated(options, ", ");
+
+    write!(w, "{name_and_anchor}({args}{between}{options})")
+}
+
 impl Textify for ScalarFunction {
     fn name() -> &'static str {
         "ScalarFunction"
     }
 
     fn textify<S: Scope, W: fmt::Write>(&self, ctx: &S, w: &mut W) -> fmt::Result {
-        let name_and_anchor =
-            NamedAnchor::lookup(ctx, ExtensionKind::Function, self.function_reference);
-        let name_and_anchor = ctx.display(&name_and_anchor);
-
-        let args = ctx.separated(&self.arguments, ", ");
-        let options = ctx.separated(&self.options, ", ");
-        let between = if self.arguments.is_empty() || self.options.is_empty() {
-            ""
-        } else {
-            ", "
-        };
+        textify_function_call_prefix(
+            self.function_reference,
+            &self.arguments,
+            &self.options,
+            ctx,
+            w,
+        )?;
 
         let output = OutputType(self.output_type.as_ref());
-        let output_type = ctx.display(&output);
-
-        write!(
-            w,
-            "{name_and_anchor}({args}{between}{options}){output_type}"
-        )?;
-        Ok(())
+        write!(w, "{}", ctx.display(&output))
     }
 }
 
@@ -532,7 +545,7 @@ impl Textify for RexType {
             RexType::Literal(literal) => literal.textify(ctx, w),
             RexType::Selection(f) => f.textify(ctx, w),
             RexType::ScalarFunction(s) => s.textify(ctx, w),
-            RexType::WindowFunction(_w) => write!(
+            RexType::WindowFunction(_f) => write!(
                 w,
                 "{}",
                 ctx.failure(PlanError::unimplemented(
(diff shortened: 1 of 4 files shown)


src/textify/expressions.rs
```diff
@@ -1,4 +1,4 @@
-use std::fmt::{self};
+use std::fmt;
```
wackywendell, line 1: 👍 , good fix, not sure why it was otherwise

src/textify/expressions.rs
```diff
@@ -402,32 +402,45 @@ impl Textify for FieldReference {
     }
 }
 
+/// Write the `name(args, options)` prefix shared by `ScalarFunction`,
+/// `AggregateFunction`, and `WindowFunction` textification.
+fn textify_function_call_prefix<S: Scope, W: fmt::Write>(
```
wackywendell: See [the previous feedback](https://github.com/DataDog/substrait-explain/pull/185/changes/63266404a85972049c1a8d5d54519ac91072cf2a#r3604792068). I’m still finding the “prefix” abstraction unclear: what makes something part of the prefix, rather than a component of the function call?

One option is to compose the pieces directly. `NamedAnchor` and `OutputType` already exist; we could add an `ArgsAndOptions(&[FunctionArgument], &[FunctionOption])` type that implements `Textify` and handles the separator between the two lists. The scalar implementation could then be roughly:

```rust
write!(w, "{name_and_anchor}({args_and_options}){output_type}")
```

The window implementation would use the same components with an additional over clause.

Alternatively, I think the fields shared by scalar, aggregate, and window functions form a reasonable data structure. It is still a little odd—it does not correspond directly to a semantic Substrait type—but it clarifies the separation between the data being formatted and the textification machinery:

```rust
struct FunctionInvocation<'a> {
    function_reference: u32,
    arguments: &'a [FunctionArgument],
    options: &'a [FunctionOption],
}
```

Implementing `Textify` for that would separate the Substrait data being formatted from the output machinery, and it would compose naturally with `ctx.display`:

```rust
write!(
    w,
    "{}{}",
    ctx.display(&invocation),
    ctx.display(&output_type),
)
```

Either approach would make the complete function textifiers read as compositions of meaningful pieces, without a separate helper whose contract is to write only the first part.

src/textify/expressions.rs
```diff
@@ -645,26 +658,16 @@ impl Textify for AggregateFunction {
...
+        )?;
 
         let output = OutputType(self.output_type.as_ref());
-        let output_type = ctx.display(&output);
-
-        write!(
-            w,
-            "{name_and_anchor}({args}{between}{options}){output_type}"
-        )
+        write!(w, "{}", ctx.display(&output))
```
wackywendell: To connect this with above - see how the previous code had the whole thing in a single `write!` call, with a template that makes it clearer how it fits together?

```
{name_and_anchor}({args}{between}{options}){output_type}
```

Maintaining separate structs here allows this kind of composition, which is easier to understand.

src/textify/values.rs
```diff
@@ -0,0 +1,358 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Arguments<'a> {
+    /// Positional arguments (e.g., a filter condition, group-bys, etc.)
+    pub positional: Vec<Value<'a>>,
+    /// Named arguments (e.g., limit=10, offset=5)
+    pub named: Vec<NamedArg<'a>>,
+    /// How this argument list is laid out. Defaults to [`ArgsLayout::Inline`];
+    /// only `Read:Virtual` opts into [`ArgsLayout::Rows`].
+    layout: ArgsLayout,
```
wackywendell: `ArgsLayout` is not used by the `Textify` implementation for `Arguments`; it is instead read by `Relation::write_header` because it affects how the whole relation header is rendered. That wasn't a great code structure to begin with, when it was in a single module; here, we're moving `Arguments` and with it `ArgsLayout` out of `rels.rs`, which makes that coupling worse.

Can we keep `ArgsLayout` separate from `Arguments` and leave it in `rels.rs`, where it is used? Alternatively, perhaps we should drop `ArgsLayout`: `Arguments` can represent inline, comma-separated arguments, and we can use a separate, relation-specific data structure for line-separated virtual rows.

wackywendell (commented):
Thanks for working through the dependency issues here: expression and relation formatting can now share value-formatting infrastructure without `expressions` depending on `rels`. The owned enum conversion is also a useful improvement.

I think there are still a few boundaries to straighten out before merging. `ArgsLayout` remains relation-specific even though it is moving with shared `Arguments`, and there's one comment about the new function-call helper that could be clearer. I'd like another pass on those points, but the overall refactoring direction looks good.

src/textify/values.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+}
+
+impl<'a> From<&'a SortField> for Value<'a> {
+    fn from(sf: &'a SortField) -> Self {
+        let field = match &sf.expr {
+            Some(expr) => match &expr.rex_type {
+                Some(substrait::proto::expression::RexType::Selection(fref)) => {
+                    if let Some(substrait::proto::expression::field_reference::ReferenceType::DirectReference(seg)) = &fref.reference_type {
+                        if let Some(substrait::proto::expression::reference_segment::ReferenceType::StructField(sf)) = &seg.reference_type {
```
wackywendell: Imports should be at the top, so that these expressions are more readable

```rust
use substrait::proto::expression::RexType;
use substrait::proto::expression::field_reference::ReferenceType as FieldReferenceType;
use substrait::proto::expression::reference_segment::ReferenceType as SegmentReferenceType;
```

```suggestion
                Some(RexType::Selection(fref)) => {
                    if let Some(FieldReferenceType::DirectReference(seg)) = &fref.reference_type {
                        if let Some(SegmentReferenceType::StructField(sf)) = &seg.reference_type {
```

wackywendell (approved):
Generally looks good! One small comment, otherwise go for it!

PR #187, opened 2026-07-21
refactor(parser): extract shared components

gord02:
## Description

This PR refactors extension arguments and function references for shared components that may require more clean deduplicated access to them. 

### Key Changes
Two related parser refactors ahead of window functions:
- Introduce a named function_reference grammar rule and a FunctionReference type (unresolved name + anchor, with resolve()), replacing the FunctionHead helper that left output type/args dangling in the caller's iterator. Scalar function parsing now composes FunctionReference + parse_argument_list + type.
- Move ParsedNamedArgs into the neutral parser::common and extract sort_direction_from_str (shared by the Sort relation), so parser modules share primitives via common rather than expressions depending on relations.

## Testing
- [x] All existing tests pass

### DIFF
--- a/src/parser/common.rs
+++ b/src/parser/common.rs
@@ -1,6 +1,8 @@
+use std::collections::HashMap;
 use std::fmt;
 
 use pest_derive::Parser as PestDeriveParser;
+use substrait::proto::sort_field::SortDirection;
 use thiserror::Error;
 
 use crate::extensions::SimpleExtensions;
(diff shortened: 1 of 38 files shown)


src/parser/expressions.rs
```diff
@@ -396,62 +396,114 @@ impl ScopedParsePair for Literal {
...
+        // Optional anchor (e.g., #1)
         let anchor = iter
             .try_pop(Rule::anchor)
             .map(|n| unwrap_single_pair(n).as_str().parse::<u32>().unwrap());
 
-        // Parse optional URN anchor (e.g., @1)
+        // Optional URN anchor (e.g., @1); currently unused.
         let _urn_anchor = iter
             .try_pop(Rule::urn_anchor)
             .map(|n| unwrap_single_pair(n).as_str().parse::<u32>().unwrap());
```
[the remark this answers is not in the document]
gord02: It is not likely a user would assign an anchor value that would overflow as that adds to the difficultly of reading the plan.

src/parser/expressions.rs
```diff
@@ -396,62 +396,114 @@ impl ScopedParsePair for Literal {
...
 
-        // Parse compound function name (required) — e.g. "equal" or "equal:any_any"
+        // Compound function name (required) — e.g. "equal" or "equal:any_any"
         let name = iter.parse_next::<CompoundName>();
 
-        // Parse optional anchor (e.g., #1)
+        // Optional anchor (e.g., #1)
         let anchor = iter
             .try_pop(Rule::anchor)
             .map(|n| unwrap_single_pair(n).as_str().parse::<u32>().unwrap());
```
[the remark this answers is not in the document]
gord02: It is not likely a user would assign an anchor value that would overflow as that adds to the difficultly of reading the plan.

src/parser/common.rs
```diff
@@ -315,6 +317,97 @@ impl Drop for RuleIter<'_> {
...
 
+/// A collection of named arguments (`name=value` pairs) extracted from a
+/// named-argument-list rule, keyed by name with duplicate-name rejection.
+///
+/// Shared by every consumer of the generic `name=value` grammar — the `Fetch`
+/// relation's `limit=`/`offset=`, and window functions' `over(...)` arguments.
+/// It lives here in `common` (rather than in `relations` or `expressions`) so
+/// that neither of those modules has to depend on the other for it.
+///
+/// The fluent API ensures all arguments are processed exactly once and none are
```
wackywendell: "fluent API"?

src/parser/expression_grammar.pest
```diff
@@ -146,15 +146,19 @@ type = { simple_type | compound_type | user_defined_type }
 
 argument_list = { "(" ~ (expression ~ (sp ~ "," ~ sp ~ expression)*)? ~ ")" }
 
-// Scalar function call:
+// An unresolved reference to a function:
```
wackywendell:
```suggestion
// A function reference:
```

'unresolved' here is a bit confusing, as resolution is not something that happens at the grammar layer.

src/parser/expressions.rs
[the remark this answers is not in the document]
wackywendell: Hm... yes, it isn't likely, but the general pattern is that applications should expect user input to be bad in any particular way, and ensure that they return errors and not crash. E.g. if we were to put this code in a running service serving 100s of queries/sec, and someone sent a request with a negative input... this panic could crash the whole server. So this is a valid concern.

That said - this pre-dates this PR, which doesn't add or change it much; this is present in the code in several places, and out of scope of this PR. I made [an issue](https://github.com/DataDog/substrait-explain/issues/201) which we should address at some point, but not needed in this PR.

src/parser/expressions.rs
```diff
@@ -580,62 +580,114 @@ impl ScopedParsePair for Literal {
...
+            self.anchor,
+            self.name.full(),
             span,
-        )?;
+        )
+    }
+}
+
+/// Parse an `argument_list` rule (`(expr, expr, ...)`) into function arguments.
+fn parse_argument_list(
```
wackywendell: Could `argument_list` have a corresponding type that implements `ScopedParsePair`, rather than being parsed through a standalone function? This conversion has the shape that trait is intended for: it maps one grammar rule to a typed value, requires extension context, and can fail while parsing its expressions.
Perhaps this could be a small `FunctionArguments(Vec<FunctionArgument>)` type. I would prefer that over implementing the trait directly for `Vec<FunctionArgument>`, since the generic container does not itself represent this particular grammar rule. That would make the extracted function-reference and argument-list components follow the same parser interface and give window-function parsing a reusable component.

src/parser/common.rs
```diff
@@ -315,6 +317,97 @@ impl Drop for RuleIter<'_> {
...
+    }
+}
+
+/// Map a sort-direction enum identifier (without the leading `&`) to a
+/// [`SortDirection`]. Shared by the `Sort` relation's `sort_field` parser and
+/// the window function's `order=` parser, which reach it from different grammar
+/// rules (`sort_direction` vs a generic `enum_value`) but accept the same set
+/// of variant names. Lives in `common` so neither `relations` nor
+/// `expressions` depends on the other for it.
+pub(crate) fn sort_direction_from_str(
```
wackywendell, line 371: Hm, I don't really have a better suggestion here right now, but... this function "smells" a bit.

1. Function comments should not list their callers ("Shared by"); the function should focus on what it does.
  - In this PR, I don't think it actually is shared? It's exported here and only used once. PRs should stand on their own, not refer to code not currently in the codebase. (Although the clarification as to why this is here is somewhat useful; that can make sense in a PR comment if you want to say something like "this looks a little odd, but is here because in the next PR I'm going to ...")
2. This function takes a `span` just for error purposes, and otherwise doesn't use it.
3. This is a bit more "taste", but it's a loose, unattached function; the code structure would be clearer if it were an `impl` of some sort, especially since its exported from this module and used in another...

All that said - 'fixing' all that would be a bit heavy; you could `impl ScopedParsePair` for `SortDirection`, but then you couldn't easily reuse that for the window function (unless it accepted both rules?). So I guess you can leave it...

wackywendell (approved):
A few relatively small comments, thanks for doing this!

PR #191, opened 2026-07-24
feat: added support for post join filter

eliot1480:
## Description

Added support for post join filter as an option field in Join relations.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [x] Pre-commit hooks pass

### DIFF
--- a/src/parser/expression_grammar.pest
+++ b/src/parser/expression_grammar.pest
@@ -339,7 +339,7 @@ fetch_named_arg_list =  { fetch_named_arg ~ (sp ~ "," ~ sp ~ fetch_named_arg)* }
 fetch_args           = _{ fetch_named_arg_list | empty }
 fetch_relation       =  { "Fetch" ~ "[" ~ fetch_args ~ sp ~ "=>" ~ sp ~ reference_list ~ "]" }
 
-// JoinRel: Join[&Inner, eq($0, $2), post_join_filter=gt($3, 0) => $0, $1, $2, $3]
+// JoinRel: Join[&Inner, eq($0, $2), post_filter=gt($3, 0) => $0, $1, $2, $3]
 // Note: Longer prefixes must come before shorter ones (e.g., "&LeftSemi" before "&Left")
 // so that the parser doesn't match "&Left" when it should match "&LeftSemi"
 join_type             = { "&Inner" | "&LeftSemi" | "&LeftAnti" | "&LeftSingle" | "&LeftMark" | "&Left" | "&RightSemi" | "&RightAnti" | "&RightSingle" | "&RightMark" | "&Right" | "&Outer" }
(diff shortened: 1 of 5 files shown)


src/parser/expression_grammar.pest
```diff
@@ -338,11 +338,12 @@ fetch_named_arg_list =  { fetch_named_arg ~ (sp ~ "," ~ sp ~ fetch_named_arg)* }
...
 fetch_relation       =  { "Fetch" ~ "[" ~ fetch_args ~ sp ~ "=>" ~ sp ~ reference_list ~ "]" }
 
-// JoinRel: Join[&Inner, eq($0, $2) => $0, $1, $2, $3]
+// JoinRel: Join[&Inner, eq($0, $2), post_join_filter=gt($3, 0) => $0, $1, $2, $3]
 // Note: Longer prefixes must come before shorter ones (e.g., "&LeftSemi" before "&Left")
 // so that the parser doesn't match "&Left" when it should match "&LeftSemi"
-join_type     = { "&Inner" | "&LeftSemi" | "&LeftAnti" | "&LeftSingle" | "&LeftMark" | "&Left" | "&RightSemi" | "&RightAnti" | "&RightSingle" | "&RightMark" | "&Right" | "&Outer" }
-join_relation = { "Join" ~ "[" ~ join_type ~ sp ~ "," ~ sp ~ expression ~ sp ~ "=>" ~ sp ~ reference_list ~ "]" }
+join_type             = { "&Inner" | "&LeftSemi" | "&LeftAnti" | "&LeftSingle" | "&LeftMark" | "&Left" | "&RightSemi" | "&RightAnti" | "&RightSingle" | "&RightMark" | "&Right" | "&Outer" }
+join_post_join_filter = { "post_join_filter" ~ sp ~ "=" ~ sp ~ expression }
```
wackywendell: [quibble] Hm... `post_join_filter` is very explicit, but also a bit long for human-readable text. We could use `where` - it's a post-join filter in SQL - but that's less clear that that maps. Perhaps `post_filter` simplifies it a bit while still being clear?

(I'm not even sure why the `post_join_filter` exists; why do we need this when we already have a JOIN condition, and you could add a `Filter` relation anyway? 🤷 it exists, so we'll support it...)

```suggestion
join_post_join_filter = { "post_filter" ~ sp ~ "=" ~ sp ~ expression }
```

I know this is a bit bikeshed-y, but part of the goal of substrait-explain is to be less verbose, and more human-readable, than the protobufs / JSON, hence my interest here.

tests/plan_roundtrip.rs
```diff
@@ -429,6 +429,42 @@ Root[id, name, amount]
     roundtrip_plan(plan);
 }
 
+#[test]
+fn test_join_relation_post_join_filter_roundtrip() {
```
wackywendell, line 433: 👍 , nice straightforward tests!

(Do we need both? 🤷  it works, we can keep it...)

wackywendell (approved):
LGTM! I have one quibble on naming, but overall looks good! Thanks, Eliot!

PR #200, opened 2026-08-05
chore: enable Clippy path and item-placement lints

wackywendell:
## Description

Enabled a couple more Clippy lints and updated code to address them.

In particular, added the `absolute_paths` lint, which warns for things like `foo:bar::baz::fizzbuzz(y)`, suggesting an import of either `baz` or `fizzbuzz` so that qualified paths are shorter.

## Commits

This is perhaps easiest reviewed by commit / commit-grouping.

- The first commit enables the lints
- the second adds the exception for generated code
- the last three commits apply the style changes

## Changes

- Enable `clippy::absolute_paths` and `clippy::items_after_statements` individually in `Cargo.toml` rather than enabling the full restriction lint group.
- Followed through in code:
  - Shorten absolute paths across the library, CLI, tests, and examples.
  - Move function-local imports and declarations to their module scope.
  - Add an exception for generated code module

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,15 @@ serde = ["substrait/serde", "dep:pbjson-types"]
 # CLI functionality with additional dependencies.
 # Includes prost-reflect for JSON parsing: both Go protojson (@type encoding) and Rust
 # pbjson (typeUrl/value encoding) are handled
-cli = ["dep:clap", "dep:serde_json", "dep:serde_yaml", "dep:anyhow", "serde", "dep:prost-reflect", "substrait/embed-descriptor"]
+cli = [
+    "dep:clap",
+    "dep:serde_json",
+    "dep:serde_yaml",
+    "dep:anyhow",
+    "serde",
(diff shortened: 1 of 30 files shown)


Cargo.toml
```diff
@@ -73,3 +81,7 @@ serde_yaml = "0.9"
 prost = "0.14"
 prost-build = "0.14"
 protox = "0.9"
+
+[lints.clippy]
+absolute_paths = "warn"
+items_after_statements = "warn"
```
edouardmulliez, line 87: do we have a check in CI that fails if there are lint warnings ? (I'm wondering if it fails on warnings or on errors)

edouardmulliez (approved):

Cargo.toml
wackywendell: Yes, we do; the `-D warnings` is for `deny warnings:

https://github.com/DataDog/substrait-explain/blob/main/.github/workflows/rust.yml#L27-L39

PR #202, opened 2026-08-07
feat(extensions)!: Add context to to_args in Explainable trait

edouardmulliez:
The goal of this change is to give access to information about the inputs when we compute the args of an extension.

I called this parameter `context` so that it can be extended with other information (ie format options) if we need to.

The concrete use case I have for this is the following one: I have an `ExtensionSingle` for which the number of output columns is equal to the number of inputs columns. When I define the extension for that `ExtensionSingle`, I extend `Explainable` and in the `to_args` method, I have no information about the input, so I cannot compute correctly the `ExtensionArgs.output_columns`.

If the current change is merged, I would use `context.inputs().emitted_column_count()` to compute the `ExtensionArgs.output_columns`.

### DIFF
--- a/src/extensions/registry.rs
+++ b/src/extensions/registry.rs
@@ -26,7 +26,8 @@
 //!
 //! ```rust
 //! use substrait_explain::extensions::{
-//!     Any, AnyConvertible, AnyRef, Explainable, ExtensionArgs, ExtensionError, ExtensionRegistry,
+//!     Any, AnyConvertible, AnyRef, Explainable, ExtensionArgs, ExtensionContext, ExtensionError,
+//!     ExtensionRegistry,
 //! };
 //!
 //! // Define a custom extension type
@@ -68,7 +69,10 @@
 //!         })
 //!     }
 //!
-//!     fn to_args(&self) -> Result<ExtensionArgs, ExtensionError> {
+//!     fn to_args(
+//!         &self,
+//!         _context: &ExtensionContext<'_>,
+//!     ) -> Result<ExtensionArgs, ExtensionError> {
 //!         let mut args = ExtensionArgs::default();
 //!         args.insert("path", self.path.clone());
 //!         Ok(args)
@@ -324,21 +328,7 @@ pub trait Explainable: Sized {
     fn from_args(args: &ExtensionArgs) -> Result<Self, ExtensionError>;
 
     /// Convert this type to extension arguments
-    fn to_args(&self) -> Result<ExtensionArgs, ExtensionError>;
-
-    /// Convert this type to extension arguments with information about its inputs.
-    ///
-    /// The default delegates to [`Explainable::to_args()`], preserving the
-    /// behavior of existing implementations. Relation extensions receive
-    /// their inputs in relation order; other extension namespaces receive an
-    /// empty input slice.
-    fn to_args_with_context(
-        &self,
-        context: &ExtensionContext<'_>,
-    ) -> Result<ExtensionArgs, ExtensionError> {
-        let _ = context;
-        self.to_args()
-    }
+    fn to_args(&self, context: &ExtensionContext<'_>) -> Result<ExtensionArgs, ExtensionError>;
 }
 
 /// Internal trait that converts between ExtensionArgs and protobuf Any messages.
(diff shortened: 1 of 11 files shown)


src/extensions/registry.rs
```diff
@@ -581,6 +652,15 @@ impl ExtensionRegistry {
...
         detail: AnyRef<'_>,
+    ) -> Result<(String, ExtensionArgs), ExtensionError> {
+        self.decode_with_type_and_context(ext_type, detail, &ExtensionContext::default())
+    }
+
+    fn decode_with_type_and_context(
+        &self,
+        ext_type: ExtensionType,
+        detail: AnyRef<'_>,
+        context: &ExtensionContext<'_>,
```
wackywendell, line 653: Why do we need `ExtensionType`?

I ask because I'd rather not expand the interface too much; that's the cost here. In terms of the benefit, I would not expect it to be much:

1. It's rather unusual for an extension to be used in more than one way, especially for that to affect how it is encoded/decoded. and seems like a bit of an anti-pattern to encode that here
2. If a user really needs that anyway, they could create two different wrapper types, and register them for different `ExtensionType`s with the different behavior built in.

wackywendell (commented):
Overall, code looks reasonable. I have two thoughts on the interface:

1. I'm not sure we need `decode_with_type_and_context`; it doesn't seem like 'good behavior' for an extension to (a) register under two different types, or (b) even if it did, to have different behavior / args depending on the type. And there are other ways to solve it without this
2. What you've done with `ExtensionInputs` and `ExtensionContext` is a good interface; the `_with_context` functions are reasonable for preserving existing behavior. That said - this is still pre-1.0, and not heavily used; I'm tempted to just add `context:&ExtensionContext<_>` to `to_args` etc. instead of duplicating the interface. Its a breaking change, but a pretty easy one to address, and might be worth it in order to keep the interface straightforward...

src/extensions/registry.rs
edouardmulliez: The fact that we need ExtensionType in the `decode_xx` methods from `ExtensionRegistry` is pre-existing to my change.

Currently, in ExtensionRegistry, we have: decode_extension_table / decode_enhancement / decode_optimization which are each calling decode_with_type with the given ExtensionType. 

We do this because we currently store the [ExtensionRegistry.type_url](https://github.com/DataDog/substrait-explain/blob/be5374cd353e93097e1f150f98e118caa3248d87/src/extensions/registry.rs#L348) in a map notably keyed by extensiontype.

I think we could:
- store `ExtensionRegistry.type_url` / `ExtensionRegistry.handlers` without extensionType in the key
- remove the decode_extension_table / decode_enhancement / decode_optimization methods, and only keep a method decode

We could also change the register_xxx API. Currently, we have register_relation/register_extension_table/register_enhancement/register_optimization. We could instead have a single register method.

These methods are public, so removing them will be a breaking change in API, but I think it's ok to do it.

If we do that change, I believe it should be done in a separate PR, either before or after my current PR.

src/textify/rels.rs
```diff
@@ -611,45 +611,61 @@ impl<'a> Relation<'a> {
...
         child_refs: Vec<Option<&'a Rel>>,
         ctx: &S,
     ) -> Self {
+        let (children, _) = Relation::convert_children(child_refs, ctx);
+        let inputs = children
+            .iter()
+            .filter_map(|child| {
+                child
+                    .as_ref()
+                    .map(|child| ExtensionInput::new(child.emitted()))
```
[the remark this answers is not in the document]
edouardmulliez: For most relations, the emit_kind is correctly handled/supported (for example, correctly supported for Project/Aggregate/Filter/...). If there are extensions where we don't correctly handle it, we should update this. But this is a change that should be done separately from the current PR.

edouardmulliez:
> Overall, code looks reasonable. I have two thoughts on the interface:
> 
> 1. I'm not sure we need `decode_with_type_and_context`; it doesn't seem like 'good behavior' for an extension to (a) register under two different types, or (b) even if it did, to have different behavior / args depending on the type. And there are other ways to solve it without this
> 2. What you've done with `ExtensionInputs` and `ExtensionContext` is a good interface; the `_with_context` functions are reasonable for preserving existing behavior. That said - this is still pre-1.0, and not heavily used; I'm tempted to just add `context:&ExtensionContext<_>` to `to_args` etc. instead of duplicating the interface. Its a breaking change, but a pretty easy one to address, and might be worth it in order to keep the interface straightforward...

@wackywendell I've updated the PR to instead directly add `context` in `to_args` arguments.

Regarding using the ExtensionType, I believe this is something that should be treated outside of this PR. Currently all the register_xx and decode_xx are by ExtensionType. We could probably simplify this instead have single register/decode methods.

src/extensions/registry.rs
wackywendell: Ah, right - I misunderstood; this is not adding `ExtensionType` to the `Explainable` interface, this is a private method of the `ExtensionRegistry`. We could perhaps clean this up, but that does seem out of scope for this PR.

wackywendell (approved):

PR #205, opened 2026-08-10
refactor(extensions)!: centralize argument extraction

wackywendell:
## Description

This PR simplifies the interface around `ArgsExtractor`, and pushes the assertions around using all arguments (`check_exhausted()`) out of user code and into the registry conversion.

```rust
fn from_args(args: &mut ArgsAccess<'_>) -> Result<Self, ExtensionError> {
    let path = args.expect_named::<&str>("path")?.to_owned();
    let batch_size = args.get_named::<i64>("batch_size")?.unwrap_or(1024);
    Ok(Self { path, batch_size })
}
```

## Changes

- Changes `Explainable::from_args` to receive `&mut ArgsAccess<'_>`.
- Removes `ExtensionArgs::extractor`, public `ArgsExtractor::check_exhausted`, and the panic-based `Drop` assertion.
- Rejects unhandled named and positional arguments after successful decoding.
- Makes `positional()` mark the complete positional list as handled.
- Keeps `output_columns()` read-only and outside exhaustion checking because output columns are relation metadata consumed independently by the structural parser.
- Preserves `Explainable::to_args(&ExtensionContext)` and context-aware relation formatting from #202.
- Rewrites the API documentation around the user-visible decoding lifecycle rather than presenting `ArgsExtractor` as an optional convenience.

## API impact

This is a breaking public API change. External `Explainable` implementations must:

- change `from_args(&ExtensionArgs)` to `from_args(&mut ArgsExtractor<'_>)`;
- use the extractor passed by the registry;
- remove explicit `check_exhausted()` calls.

There is intentionally no new public direct-conversion helper. The registry owns the checked decoding boundary.

## Validation

- `cargo test --all-features`
- `just check`
- Independent review found no actionable issues

### DIFF
--- a/API.md
+++ b/API.md
@@ -101,7 +101,7 @@ Substrait has two extension mechanisms:
 
 - **Simple extensions** declare extension functions, types, and type variations.
   `substrait-explain` reads and writes these declarations in the `===
-  Extensions` section and uses them to resolve anchors while parsing and
+Extensions` section and uses them to resolve anchors while parsing and
   formatting expressions and types. No YAML files are read; substrait-explain
   relies on the protobufs / text format itself for function names, and does not
   validate type signatures exist / match extensions.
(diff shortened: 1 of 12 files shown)


edouardmulliez (approved):
good refacto!

PR #207, opened 2026-08-12
feat: added support for setting common defaults explicitly

eliot1480:
## Description

Every relation `substrait-explain` produces now carries an explicit `RelCommon.emit_kind`, rather than leaving `common` unset when the emit is a plain pass-through.

`emit_kind` is a protobuf `oneof`, so an unset one has no default value on the wire. Unlike a scalar enum, absence doesn't decode to a zero value, so a consumer reading a relation without it has to infer/guess that the relation passes its columns through unchanged, and writing `direct: {}` eliminates any risk out of the plan.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Example update
- [ ] Other (please describe)

## Testing

- [x] All existing tests pass

## Checklist

- [ ] No breaking changes (or breaking changes documented)
- [x] Pre-commit hooks pass

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -569,6 +569,13 @@ RelationName[arguments, named_arguments => columns]
 
 The exact structure varies by relation type, but all follow this basic pattern.
 
+#### Emit
+
+Every relation this crate parses carries an explicit `RelCommon.emit_kind`:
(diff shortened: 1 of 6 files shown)


wackywendell (approved):
Good catch! There's a formatting issue, so fix that and we're good to merge 👍

PR #210, opened 2026-08-17
feat: Support expressions inside Sort relations

edouardmulliez:
In substrait, in a `SortField`, we can use an expression. Up to now, we were only supporting direct references in the SortField.

This change adds support for `SortField` which are using an expression which is not a direct field reference.

For example:
```
Root[a, b]
  Sort[(add($0, $1):i32, &AscNullsFirst) => $0, $1]
    Read[table => a:i32, b:i32]"#;
```

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -1088,22 +1088,22 @@ Root[result]
 
 ### Sort Relation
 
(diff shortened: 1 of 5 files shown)


wackywendell (approved):
LGTM! This is a great change, shows the abstractions working well. Godo find, much appreciated!

PR #213, opened 2026-08-19
fix: making version non-default

gord02:
## Description

The version in the plan proto is currently always printed. This information ins't always necessary to view when printing a plan so we should print this primarily in verbose mode. 

## Type of Change
- [x] Bug fix

## Testing
- [x] All existing tests pass

### DIFF
--- a/API.md
+++ b/API.md
@@ -364,6 +364,7 @@ cat plan.substrait | substrait-explain convert -f text -t json > plan.json
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
 - `--show-literal-types` - Show type annotations on literals
+- `--show-plan-version` - Show the plan's Substrait version
 - `--verbose` - Show detailed progress information
 
 #### Validate Command
@@ -385,6 +386,7 @@ substrait-explain validate -i plan.substrait --verbose
 
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
+- `--show-plan-version` - Show the plan's Substrait version
 - `--verbose` - Show detailed progress information
 
 ### Examples
--- a/API.md
+++ b/API.md
@@ -363,8 +363,7 @@ cat plan.substrait | substrait-explain convert -f text -t json > plan.json
 - `-t, --to <FORMAT>` - Output format (default: text)
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
-- `--show-literal-types` - Show type annotations on literals
-- `--show-plan-version` - Show the plan's Substrait version
+- `--detailed` - Show more detail on plans, including type annotations and plan version
 - `--verbose` - Show detailed progress information
 
 #### Validate Command
@@ -386,7 +385,6 @@ substrait-explain validate -i plan.substrait --verbose
 
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
-- `--show-plan-version` - Show the plan's Substrait version
 - `--verbose` - Show detailed progress information
 
 ### Examples
@@ -396,8 +394,8 @@ substrait-explain validate -i plan.substrait --verbose
 substrait-explain validate -i example-plans/basic.substrait
 substrait-explain validate -i example-plans/simple.substrait
 
-# Convert with verbose output and type information
-substrait-explain convert -f text -t json --show-literal-types --verbose -i example-plans/basic.substrait
+# Convert with verbose output and full plan detail
+substrait-explain convert -f text -t json --detailed --verbose -i example-plans/basic.substrait
 
 # Roundtrip test: text → protobuf → text
 substrait-explain convert -f text -t protobuf -i plan.substrait -o plan.pb
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -125,8 +125,6 @@ version at all:
 === Version null
 ```
 
-Output leaves the section out unless `OutputOptions::show_version` asks for it.
-
 ```rust
 # use substrait_explain::Parser;
 #
@@ -145,32 +143,16 @@ Root[result]
 ```
 
 ```rust
-# use substrait_explain::{Parser, default_plan_version};
+# use substrait_explain::Parser;
 #
-# let null_version = r#"
+# let plan_text = r#"
 === Version null
 === Plan
 Root[result]
   Read[orders => quantity:i32?]
 # "#;
 #
-# let plan = Parser::parse(null_version).unwrap();
-# assert!(plan.version.is_none());
-#
-# let no_section = r#"
-# === Plan
-# Root[result]
-#   Read[orders => quantity:i32?]
-# "#;
-#
-# let plan = Parser::parse(no_section).unwrap();
-# assert_eq!(plan.version, Some(default_plan_version()));
-#
-# // nothing and `null` are different inputs, and stay different.
-# assert_ne!(
-#     Parser::parse(null_version).unwrap().version,
-#     Parser::parse(no_section).unwrap().version
-# );
+# Parser::parse(plan_text).unwrap();
 ```
 
 #### Extension format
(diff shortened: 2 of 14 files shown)


src/parser/structural.rs
```diff
@@ -917,6 +929,9 @@ impl<'a> Parser<'a> {
     /// - A plan section starting with "=== Plan"
     /// - Indented relation definitions
     ///
+    /// A document with no version(`=== Version null`) section gets
+    /// [`default_plan_version`](crate::default_plan_version);
+    ///
```
```suggestion
```

This comment doesn't fit here, it's the wrong level of detail here.

Also, this seems wrong anyway:
- `=== Version null` in the text should translate to _unset_ `Version` in the protobuf
- A missing `Version` section should translate to the 'default' plan version

src/parser/structural.rs
```diff
@@ -946,7 +961,7 @@ impl<'a> Parser<'a> {
             line_no: 1,
             state: State::Initial,
             cursor: None,
-            version: None,
+            version: Some(default_plan_version()),
```
wackywendell, line 969:
```suggestion
            // We set the version to our substrait-explain default.
            // An explicitly set version (either null or with a 
            // version number / attributes) will override this.
            // `None` here means 'explicitly unset'.
            version: Some(default_plan_version()),
```

Given that its unusual to set something to `Some` by default and then override it to `None`, this could use explanation.

src/cli.rs
```diff
@@ -101,13 +102,14 @@ impl Cli {
             Commands::Validate {
                 input,
                 output,
+                show_plan_version,
```
wackywendell: Why are we adding this specifically to `Validate`? That seems odd...

src/cli.rs
```diff
@@ -78,13 +78,14 @@ impl Cli {
                 from,
                 to,
                 show_literal_types,
+                show_plan_version,
```
wackywendell: I think instead of adding these as separate options, we should add a `--detailed` flag which uses `OutputOptions::verbose()`. I'm not sure we need specific flags for every option; we don't have them now.

src/cli.rs
```diff
@@ -292,6 +311,9 @@ pub enum Commands {
         /// Output file (use - for stdout)
         #[arg(short, long, default_value = "-")]
         output: String,
+        /// Show the plan's Substrait version (text output only)
+        #[arg(long)]
+        show_plan_version: bool,
```
wackywendell: Why is this an option for `Validate`? What would it mean to enable / disable it?

Without good reason for adding it, I don't think we should.

src/lib.rs
```diff
@@ -41,6 +41,9 @@ use textify::plan::PlanWriter;
 /// - A plan section starting with "=== Plan"
 /// - Indented relation definitions
 ///
+/// A document with no version section gets [`default_plan_version`];
+/// `=== Version null` gets no version.
+///
```
wackywendell: 🤔 As above, this also shouldn't be here. Side note... why is this comment duplicating what's in `Parser::parse()`?

```suggestion
```

src/parser/structural.rs
```diff
@@ -895,7 +906,8 @@ pub struct Parser<'a> {
     /// [`next_chunk`](Self::next_chunk). `None` before parsing starts and once
     /// the input is exhausted.
     cursor: Option<ChunkCursor<'a>>,
-    /// The plan version from an optional `=== Version` section
+    /// The plan version. Starts as [`default_plan_version`], which a
```
wackywendell, line 909: 👍

tests/json_parsing.rs
```diff
@@ -102,6 +102,16 @@ fn test_text_path() {
         .parse_plan(PLAN_TEXT)
         .expect("failed to parse text plan");
 
+    // The text fixture has no `=== Version` section, so parsing fills in the
```
wackywendell: How about we add the `Version` to the fixture instead, instead of working around it?

src/cli.rs
```diff
@@ -143,18 +146,27 @@ impl Cli {
...
             }
 
-            Commands::Validate { verbose, .. } => {
-                self.run_validate_with_io(reader, writer, *verbose, registry)
-            }
+            Commands::Validate {
+                show_plan_version,
+                verbose,
+                ..
+            } => self.run_validate_with_io(reader, writer, *show_plan_version, *verbose, registry),
```
wackywendell: Firstly, I think we should just drop `show_plan_version` as an option to `Validate`. 

Secondly, even if there was a good reason to have this, we should add `options` here, not add each option one by one.

tests/plan_roundtrip.rs
```diff
@@ -1121,14 +1125,31 @@ Root[sum]
     roundtrip_plan(plan);
 }
 
+/// Round-trip `input` with the version section at the given visibility, leaving
+/// every other option at its default.
+fn roundtrip_showing_version(input: &str, show_version: Visibility) {
```
wackywendell: This is oddly specific. May I suggest:

```rust
pub fn roundtrip_plan(input: &str) {
    roundtrip_plan_with_options(input, &OutputOptions::default());
}

pub fn roundtrip_with_options(input: &str, options: &OutputOptions) {
    …
}
```

Then in your tests:

```rust
let options = OutputOptions {
    show_version: Visibility::Required,
    ..OutputOptions::default()
};

roundtrip_plan_with_options(plan, &options);
```

That's a more generally useful function, and both arguments (`plan` and `options` are at the same abstraction level.

wackywendell: Alternatively, this interface is getting complex enough now it's probably worth framing our data structures to fit:

```rust
struct Roundtrip<'a> {
    canonical: &'a str,
    equivalent_inputs: Vec<&'a str>,
    options: OutputOptions,
}
```

Then you can use it like so:

```rust
// Simple roundtrip
Roundtrip::new(plan).assert();

// with options
Roundtrip::new(plan)
    .with_options(OutputOptions {
        show_version: Visibility::Always,
        ..OutputOptions::default()
    })
    .assert();

// with a canonical and equivalent plan
Roundtrip::new(canonical)
    .also_accepts(equivalent)
    .assert();
```

This is a non-blocking suggestion - feel free to just do the `roundtrip_with_options` suggested above.

API.md
```diff
@@ -364,6 +364,7 @@ cat plan.substrait | substrait-explain convert -f text -t json > plan.json
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
 - `--show-literal-types` - Show type annotations on literals
+- `--show-plan-version` - Show the plan's Substrait version
```
wackywendell:
```suggestion
- `--detailed` - Show more detail on plans, including type annotations and plan version
```

Let's make this one option rather than two.

API.md
```diff
@@ -385,6 +386,7 @@ substrait-explain validate -i plan.substrait --verbose
 
 - `-i, --input <FILE>` - Input file (default: stdin)
 - `-o, --output <FILE>` - Output file (default: stdout)
+- `--show-plan-version` - Show the plan's Substrait version
```
wackywendell:
```suggestion
```

Let's drop this, it doesn't make much sense as part of `validate`

GRAMMAR.md
```diff
@@ -136,6 +144,35 @@ Root[result]
...
+# "#;
+#
+# let plan = Parser::parse(no_section).unwrap();
+# assert_eq!(plan.version, Some(default_plan_version()));
+#
+# // nothing and `null` are different inputs, and stay different.
+# assert_ne!(
+#     Parser::parse(null_version).unwrap().version,
+#     Parser::parse(no_section).unwrap().version
+# );
```
```suggestion
```

You don't need a full test here asserting lots of things - the tests in `GRAMMAR.md` should merely assert that the example plan is accepted as valid input.

src/cli.rs
gord02: Since validate does a roundtrip the flag's purpose was to retain the version if it was added in the input. I can drop just use the --detailed flag.

src/textify/foundation.rs
```diff
@@ -61,6 +70,7 @@ pub struct OutputOptions {
 impl Default for OutputOptions {
     fn default() -> Self {
         Self {
+            show_version: Visibility::Never,
```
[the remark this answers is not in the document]
gord02: This is a good catch. With show_version: Never as the default, the section was omitted for every plan, including ones that arrived from JSON/protobuf carrying a real version. re-parsing that output added default_plan_version(), so an ordinary conversion overwrote the original version. Using Visibility::Required as the default writes the version unless it's empty or the default_plan_version() the parser adds. the two cases now re-parse identically.

